PR #10, opened 2016-06-21
Set window long ptr 64bit compatibility

nicolas-cellier-aka-nice:
The GetWindowLong and SetWindowLong(W) functions work for 32bits pointer only...
Use GetWindowLongPtr instead of GetWindowLong (same for Set) so as to be 64bits compatible.

See Microsoft documentation, i.e. https://msdn.microsoft.com/en-us/library/windows/desktop/ms633585%28v=vs.85%29.aspx

This is is 2nd pull request, with corrected GWL_STYLE and GWL_EXSTYLE which must not be renamed to GWLP_ unlike other constants.


bertfreudenberg:
There is no need to make new pull requests all the time. Just keep pushing to your branch, the commits will be added to this pull request automatically. When merging this PR, it will take the tip of the branch.

fniephaus:
`continuous-integration/appveyor/pr` fails because of a deployment problem. It looks like @timfel is currently working on deployments.

PR #11, opened 2016-06-21
fix deployment to only run on master and Cog branch, and not for pull requests

timfel:


timfel:
@fniephaus: can you check my bash foo?

PR #12, opened 2016-06-22
Fix ARM builds for Newspeak.

rmacnak:
- Don't hardcode a Raspian configuration, which breaks ARMv7 Linux.
 - Disable fast bitblt as it is broken when assembled by ARMv7 gcc.
 - Don't link with libasound as Newspeak VM have no sound plugin.

### DIFF
--- a/build.linux32ARM/newspeak.cog.spur/build.assert/mvm
+++ b/build.linux32ARM/newspeak.cog.spur/build.assert/mvm
@@ -18,11 +18,11 @@ test -f plugins.ext || (test -f ../plugins.ext && cp -p ../plugins.ext . || cp -
 test -f config.h || ../../../platforms/unix/config/configure --without-npsqueak \
 		--without-vm-display-fbdev --with-vmversion=5.0 \
 		--with-src=nsspursrc \
-		--without-vm-display-fbdev --without-npsqueak --enable-fast-bitblt \
-	CC="gcc -march=armv6 -mfpu=vfp -mfloat-abi=hard" \
+		--without-vm-display-fbdev --without-npsqueak \
+	CC=gcc \
 	CXX=g++ \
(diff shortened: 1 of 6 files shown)


timfel:
Hi, two questions:

1) should the mvm scripts be symlinks of one another? or are there significant differences? Just seems like a maintenance headache (not only in the newspeak build dirs, in all others, too)

2) I think we're building and uploading armv6 on Travis. Should we make this armv7?

eliotmiranda:
Hi Ryan,  Hi Tim,

    I can't accept your change Ryan, because while it fixes builds for you, it breaks builds for Tim Rowledge and NuScratch, which have 2.5 million users.  I'm not sure what's best here.  We have to split the v6 builds, which are needed for Pi, from the v7 builds which are useful elsewhere.  So that means having two sets of travis builds for ARM32.  But how should we rename?

a) leave build.linux32ARM as the "v6" build, and add build.linux32ARMv7
b) change build.linux32ARM as in your commit and add build.linux32ARMv6 (since the mvm would default and not specify)
c) rename build.linux32ARM to build.linux32ARMv6 and add build.linux32ARMv7

b) or c) are attractive, but both need to be done in the same commit, and then separately someone needs to update travis, and get the nomenclature right.  I'd like to see the "old" build products clearly marked with either v6 or RPi.  Then the new products could be labeled as v7 or...?

P.S. How do we get notified of this discussion in email?  I'm visiting the site increasingly often, but it would be nice if I could get notifications, and of only this that I'm interested in.  Things could get v busy around here...

rmacnak:
Hi Eliot,

You'll notice I made these changes only for the Newspeak builds, not the Squeak ones, so Scratch should be unaffected.  I'll defer on whether you'd rather split the build directory.

You should get notification emails for any issue or pull request you start or comment on. You can also click the subscribe button off to right if you're interested in other issues.

eliotmiranda:
Better handled IMO by duplicating the build.linux32ARM directory into versions for v6 (pi) and v7 (everything else).

PR #22, opened 2016-07-05
Prepare Cocoa VM for High-DPI

krono:
- Works for both OpenGL and CoreGraphics backends (Why do we have
   both?)
 - Needs new abstractions for 'squeak-screen-size' and
   'squeak-mouse-position' in both backends/view. Is provided.


krono:
Note that High-DPI is disabled by default in the `.plist`

PR #24, opened 2016-07-07
Make scripts fail-stop.

rmacnak:


fniephaus:
I'd prefer a `set -e` right after the shebang.
This has also caused a few builds to fail for some reason.

eliotmiranda:
Ryan, Fabio, can I leave this to you two to agree upon?  Seems like a great change to me, but Fabio you have a niggle.

fniephaus:
Hi Ryan,
I've restarted affected builds to rule out that the builds were just flaky, but it looks like this PR definitely breaks a few builds.
Could you look into this please?

fniephaus:
Forgot to mention: all failing builds are Newspeak builds on Linux...

fniephaus:
It looks like it fails because it can't find the sources file (see [this script](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/master/build.linux32x86/editnewspeakinstall.sh)).
@eliotmiranda do Newspeak builds really need a sources file?

fniephaus:
Merged: https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/3e2b3e429307bff51036ebdf551261e715af337b (had to rebase, so GitHub didn't pick up the merge automatically).

PR #25, opened 2016-07-11
Add ioScreenScaleFactor.

krono:
- if in doubt, returns 1.0.
 - enables high-res support for Windows and Mac
 - Is support for VMMaker.oscog-topa.1900


platforms/iOS/vm/OSX/sqSqueakMainApplication+screen.h
```diff
@@ -32,6 +32,7 @@
 #import "sqMacHostWindow.h"
 
 @interface sqSqueakMainApplication (screen)
+- (sqInt) ioScreenScaleFactor;
```
fniephaus: I thought this was a `double`.

krono: should be…

PR #29, opened 2016-07-18
Map X windows properly when opening or reopening the display, as requ…

dtlewis290:
…ired for #forkSqueak to open a new X display window for a child Squeak image.


fniephaus:
I had to restart two builds which failed because of some unrelated problem, now everything's green.

krono:
Looks good to me.

PR #31, opened 2016-07-20
Revise the win32 makefiles for a modern cygwin

nicolas-cellier-aka-nice:
Hacks in .travis_build.sh have been incorporated in the makefiles and can now be removed.
The toolchain config has been separated in its own file (Makefile.tools).

Note that the Bochs plugins production has not been tested by the author
It's also not yet tested by appveyor (the .travis_build.sh remove them from plugins.ext)


fniephaus:
LGTM

PR #38, opened 2016-08-17
Changed scripts so they are explicitly interpreted by bash and not the default shell

lauraperezcerrato:
Dash is the default shell for Debian and other Linux distributions, and it's unable to run most of the scripts provided to build a VMMaker trunk image.


fniephaus:
Thanks! 😃

PR #45, opened 2016-08-29
Modified the mechanism that validated the Display being a 'sane' object since it wans't appropriate for Spur's object format field

lauraperezcerrato:
For such validation to be correct, one would have to shift the displayObject oop 24 bits, not 8 (see SpurMemoryManager>>#formatShift) but displayObject() is called and its result is used without validation on other parts of the codebase, so it seems simpler to get rid of it altogether.

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -1332,12 +1332,13 @@ static sqInt display_clipboardReadIntoAt(sqInt count, sqInt byteArrayIndex, sqIn
 static void redrawDisplay(int l, int r, int t, int b)
 {
   extern sqInt displayObject(void);
+  extern sqInt isPointers(sqInt);
   extern sqInt lengthOf(sqInt);
   extern sqInt fetchPointerofObject(sqInt, sqInt);
 
   sqInt displayObj= displayObject();
 
-  if (lengthOf(displayObj) >= 4)
+  if (isPointers(displayObj) && lengthOf(displayObj) >= 4)
     {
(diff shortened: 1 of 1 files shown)


eliotmiranda:
The file should use isPointers from sqVirtualMachine.h.  Look at e.g.
$ egrep "VirtualMachine|interpreterProxy" platforms/unix/vm-display-X11/sqUnixOp
enGL.X11
#include "sqVirtualMachine.h"
extern struct VirtualMachine *interpreterProxy;
  fn= interpreterProxy->ioLoadFunctionFrom("ioGetDisplay", NULL);
  fn= interpreterProxy->ioLoadFunctionFrom("ioGetWindow", NULL);

So change
static void redrawDisplay(int l, int r, int t, int b)
{
  extern sqInt displayObject(void);
  extern sqInt lengthOf(sqInt);
  extern sqInt fetchPointerofObject(sqInt, sqInt);

  sqInt displayObj= displayObject();

  if ((((((unsigned)(oopAt(displayObj))) >> 8) & 15) <= 4)
      && ((lengthOf(displayObj)) >= 4))
    {
to
static void redrawDisplay(int l, int r, int t, int b)
{
  extern sqInt displayObject(void);
  extern sqInt isPointers(sqInt);
  extern sqInt lengthOf(sqInt);
  extern sqInt fetchPointerofObject(sqInt, sqInt);

  sqInt displayObj= displayObject();

  if (isPointers(displayObj) && lengthOf(displayObj) >= 4)
    {

More correctly it would call these functions through interpreterProxy, but the style above is faster.

lauraperezcerrato:
Thanks for the clarification!

PR #46, opened 2016-09-05
Added quicksort implementation to allow sorting FloatArrays via primitive

lauraperezcerrato:


lauraperezcerrato:
Tyding history and submitting again afterwards

StephanEggermont:
This quicksort can run out of stack space, and will be very slow with an already sorted array.
You might want to consider doing median of three pivot selection, switching to insertion sort for small subarrays (<10 or so) and  sorting smallest subarray first

lauraperezcerrato:
Added a few of the suggested improvements, let's folow the discussion on the updated pull request https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/47

PR #47, opened 2016-09-05
Implementation of quicksort such as to sort FloatArrays faster using primitives

lauraperezcerrato:
I'd like to merge the C code before pushing the primitive implementation to VMMaker

### DIFF
--- a/platforms/Cross/plugins/FloatArrayPlugin/sqFloatArrayPlugin.c
+++ b/platforms/Cross/plugins/FloatArrayPlugin/sqFloatArrayPlugin.c
@@ -1,0 +1,29 @@
+void quicksort(float* array, int from, int to) {
+    if (from < to) {
+        int p = partition(array, from, to);
+        quicksort(array, from, p);
+        quicksort(array, p + 1, to);
+    }
+    return;
+}
+
+int partition(float* array, int from, int to) {
+    float pivot = array[from];
+    int i = from - 1;
+    int j = to + 1;
+    while (-1) {
+        do {
+            i += 1;
+        } while (array[i] < pivot);
+        do {
+            j -= 1;
+        } while (array[j] > pivot);
+        if (i >= j) {
+            return j;
+        }
+        float a = array[i];
+        float b = array[j];
+        array[i] = b;
+        array[j] = a;
+    }
+}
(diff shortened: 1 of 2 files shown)


platforms/Cross/plugins/FloatArrayPlugin/sqFloatArrayPlugin.c
```diff
@@ -0,0 +1,93 @@
...
+
+int partition(float* array, int from, int to) {
+    float possiblePivots[] = { array[from], array[(to - from) / 2 + from], array[to] };
+    float pivot;
+    pivot = possiblePivots[0] < possiblePivots[2] ? possiblePivots[0] : possiblePivots[2];
+    pivot = pivot > possiblePivots[1] ? pivot : possiblePivots[1];
+    
+    int i = from - 1;
+    int j = to + 1;
+    while (-1) {
```
fniephaus, line 64: why do you use `-1` here?

platforms/Cross/plugins/FloatArrayPlugin/sqFloatArrayPlugin.c
```diff
@@ -0,0 +1,93 @@
...
+        array[j] = a;
+    }
+}
+
+void insort(float* array, int from, int to) {
+    int i;
+    for (i = from; i <= to; i++) {
+        int j = i;
+        while (j > 0 && array[j - 1] > array[j]) {
+            float a = array[j - 1];
```
fniephaus, line 86: wouldn't it be better to define `a` and `b` before the while loop, so that you can use `j > 0 && a > b` for the while condition?

krono: Won't work, as `a` and `b` depend o `j`.
This way, the two are localized to the inner loop.

Here's another variant:

``` C
void insort(float* array, int from, int to) {
 for (int i = from; i <= to; i++) {
    for (int j = i; j > 0 && array[j - 1] > array[j]; j--) {
      float xchange = array[j];
      array[j] = array[j - 1];
      array[j - 1] = xchange;
    }
  }
}
```

krono:
Apart from the small technical detail already commented, I don't think that fast sort is the job of the FloatArrayPlugin, but rather the Image's or the JIT's responsibility. 

Also, that would avoid subtile semantic deviation.

PR #52, opened 2016-09-12
Merge with Pharo (1). OSX 32/64 bits

estebanlm:
So... this is a first HUGE change :)
This change includes (not limited to):

- Correct Makefiles for building `build.macos32x86/pharo.cog.spur`
- Correct Makefiles for building `build.macos64x64/pharo.cog.spur`
- Merge back some platform files from Pharo branch 
    - they should all be correctly delimited with `PharoVM` definition
    - it is not finished, more changes will come

So, I hope this does not break anything... After this I will move to merge linux branches.


estebanlm:
mmm... it breaks something in squeak (weird, I was sure I didn't change anything)
I will retract and review.

.travis_build.sh
```diff
@@ -105,7 +105,7 @@ case "$PLATFORM" in
     pushd "${build_directory}"
 
     travis_fold start build_vm "Building OpenSmalltalk VM..."
-    ./mvm -f
+    make
```
nicolas-cellier-aka-nice: I don't know the answer, but i wonder, is it still necessary?
From my understanding, there's been 2 changes in this area:
- https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/46d8e8cd25d9b92ffd21ce21dcdeb2406b18c804
- https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/3e2b3e429307bff51036ebdf551261e715af337b
the LOGD artifacts might be of importance (we could track non regresion of warnings for example) and it won't be produced by direct call to make.

.travis_build.sh
```diff
@@ -128,7 +128,7 @@ case "$PLATFORM" in
     pushd "${build_directory}"
     # remove bochs plugins
     sed -i 's/Bochs.* //g' plugins.ext
-    make
+	./mvm -f
```
nicolas-cellier-aka-nice: I like this one better, but does it work better w.r.t. error exit?

estebanlm: Hi, @nicolas-cellier-aka-nice 
this is not like that since a lot... my current version of this file reads like this: 

```bash -e ./mvm -f```

what are you checking?

nicolas-cellier-aka-nice: Big pull requests are scary, it's sort of take it or leave it as a whole...

To mitigate that, I was trying to review the changes you submitted thru github web interface:
https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/52
Unfortunately, it's not possible (too many changes, diffs are not displayed).

So I tried to pick and review intermediate changes. But then there's the risk to review obsolescent things...
Especially, if there has been several cross-merging (rather than rebase).
Unless there's a massive rewrite this also is a dead end (so forget about above remarks).

Unfortunately, we're back to take it or leave it initial status...

PR #56, opened 2016-09-20
Make Pharo use "standard" sources (spursrc, spur64src and src/plugins)

estebanlm:
- best than needing a regeneration each time (for osvm... we still make the generation but we don't care :P)
- so we can remove (and we do it) pharosrc and pharo64src


estebanlm:
yes, lots of file changes but this is just the generation of sources for  and removal of others :)

estebanlm:
wow, I really screw this :)
Something with the FilePlugin... I will retract and retry later.

PR #63, opened 2016-09-24
FilePlugin: Remove misleading check for end of line

zecke:
When operating on stdin sqFileReadIntoAt will read one zero or one
byte from the FILE*. The loop will exit when bytesRead is not zero
and this will happen when a character was read. There is no reason
to exit the loop differently for \r or \n.

### DIFF
--- a/FilePlugin/sqFilePluginBasicPrims.c
+++ b/FilePlugin/sqFilePluginBasicPrims.c
@@ -395,12 +395,8 @@ sqFileReadIntoAt(SQFile *f, size_t count, char* byteArrayIndex, size_t startInde
 		bytesRead = 0;
 		do {
 			clearerr(file);
-			if (fread(dst, 1, 1, file) == 1) {
+			if (fread(dst, 1, 1, file) == 1)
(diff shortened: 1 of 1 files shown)


nicolas-cellier-aka-nice:
For some reason, this PR was duplicated in https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/64

zecke:
Not duplicate but different branch. The README.md made me think that this should be updated too.

```Each vm directory contains support for the core VM.  Each plugin directory
contains run-time and build-time support for various plugins.  The following
directories are subtrees that are shared with the old Squeak interpreter source:

 - platforms/Cross/plugins
   - https://github.com/OpenSmalltalk/vm/tree/platform/Cross/plugins
```

nicolas-cellier-aka-nice:
Ah OK, I see it now.
This branch is abandon-ware, there haven't been a single commit here since svn -> git move...
So I think it's reasonable to close this PR.
If ever legacy interpreter is revived on git, there will be another way than thru a way behind branch...

PR #64, opened 2016-09-24
FilePlugin: Remove misleading check for end of line

zecke:
When operating on stdin sqFileReadIntoAt will read one zero or one
byte from the FILE*. The loop will exit when bytesRead is not zero
and this will happen when a character was read. There is no reason
to exit the loop differently for \r or \n.

### DIFF
--- a/platforms/Cross/plugins/FilePlugin/sqFilePluginBasicPrims.c
+++ b/platforms/Cross/plugins/FilePlugin/sqFilePluginBasicPrims.c
@@ -396,12 +396,8 @@ sqFileReadIntoAt(SQFile *f, size_t count, char* byteArrayIndex, size_t startInde
 		bytesRead = 0;
 		do {
 			clearerr(file);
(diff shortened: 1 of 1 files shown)


nicolas-cellier-aka-nice:
👍 These changes are correct.
However, reading the mailing list thread, it seems that the intention was to read more than 1 character eventually, up to end of line (the primitive should read up to count characters).
http://lists.squeakfoundation.org/pipermail/vm-dev/2016-September/023232.html

So if we remove the cr/lf hack which is quite useless now, we also erase the intention...
Is it possible to open an issue so that we can come back to it later?

nicolas-cellier-aka-nice:
Compliantly to my own request, I opened https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/150, so that we do not forget why the cr/lf scanning was here.

It would be cool to get a little more contribution when the integrators emit such request, otherwise the PR might be rotting for so looong (10 month already is like an eternity).

PR #71, opened 2016-10-04
replaces most sprintf with snprintf

santana:
sprintf is inherently unsafe, easy to misuse. snprintf makes more robust code, it is harder to misuse.

### DIFF
--- a/platforms/unix/vm/sqUnixExternalPrims.c
+++ b/platforms/unix/vm/sqUnixExternalPrims.c
@@ -161,7 +161,8 @@ tryLoadModule(char *in, char *name)
       continue;
     }
   }
-  sprintf(out, "/" MODULE_PREFIX "%s" MODULE_SUFFIX, name);
+  snprintf(out, sizeof(path) - (out - path),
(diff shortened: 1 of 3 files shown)


platforms/unix/vm/sqUnixExternalPrims.c
```diff
@@ -202,11 +203,7 @@ void *ioLoadModule(char *pluginName)
 
   /* try dlopen()ing LIBRARY_PREFIX<name>LIBRARY_SUFFIX searching only the default locations modulo LD_LIBRARY_PATH et al */
 
-# if defined(HAVE_SNPRINTF)
```
krono, line 205: Why putting away the `if defined`?

santana: because any modern Unix already includes snprintf. All operating systems under the platforms directory have snprintf.

krono: It's specified since SUSv2 and C99.
Most of the code is not C99-Robust yet :)

krono: Anyway, go ahead :)

PR #72, opened 2016-10-04
snprintf

santana:
sprintf is inherently unsafe, easy to misuse. snprintf makes more robust code, it is harder to misuse.


platforms/unix/vm/sqUnixExternalPrims.c
```diff
@@ -452,11 +441,7 @@ ioFindExternalFunctionIn(char *lookupName, void *moduleHandle)
   if (fn && accessorDepthPtr) {
 	signed char *accessorDepthVarPtr;
 
-# ifdef HAVE_SNPRINTF
 	snprintf(buf+strlen(buf), sizeof(buf), "AccessorDepth");
```
bencoman: btw, shouldn't this existing code be...
  snprintf(buf+strlen(buf), sizeof(buf) - strlen(buf), "AccessorDepth");

platforms/unix/vm/sqUnixExternalPrims.c
```diff
@@ -202,11 +203,7 @@ void *ioLoadModule(char *pluginName)
 
   /* try dlopen()ing LIBRARY_PREFIX<name>LIBRARY_SUFFIX searching only the default locations modulo LD_LIBRARY_PATH et al */
 
-# if defined(HAVE_SNPRINTF)
```
bencoman, line 205: hi Gerardo,  
A good initiative, but these HAVE_SNPRINTF checks indicate some platforms don't support snprintf().  However this code is from 2010 and maybe there is broader support now. But in particular MSVC didn't support snprintf() until 2015. I do see Nicolas reports successful builds using MSVC2015 - so I guess that would become a the minimum requirement for native Windows.  

Any platforms missing snprintf() might pick up a portable version [1] from instead #defining snprintf() as sprintf()...
[1] https://github.com/search?utf8=%E2%9C%93&q=snprintf

santana: I agree with you on using a portable snprintf where it is not available. That means two things:
1. Those HAVE_SNPRINTF would go away anyways, because instead of being used everywhere in the source code, it will be used in a single place: the Makefiles, for conditionally linking the portable version.
2. I think that platforms we care about support snprintf, so there are no reasons for working around in the code for the lack of snprintf (and even more so when, as you pointed out, we can use a portable version)

platforms/unix/vm/sqUnixExternalPrims.c
santana: wow, yes,  you have just spotted a current bug :)

PR #73, opened 2016-10-04
add pharo 64bits linux and some minor enhancements (2nd attempt)

estebanlm:
this add 64bits for linux but also adapts command line to accept "pharo format" command (double dash instead one: --) 
also it includes a couple of minor PharoVM changes (like XInitThreads)


build.linux32x86/third-party/libgit2/Makefile.lib
```diff
@@ -0,0 +1,52 @@
...
+$(LIBGIT2ARCHIVE): 
+	curl -o $(LIBGIT2ARCHIVE) -L $(LIBGIT2URL) 
+
+$(LIBGIT2DIR)/CMakeLists.txt: $(LIBGIT2ARCHIVE)
+	tar x -f $(LIBGIT2ARCHIVE) -C $(THIRDPARTYDIR)
+	touch $(LIBGIT2DIR)/CMakeLists.txt
+
+# NOTE: I'm adding this ugly hack of "sed..." because CMake does not find properly libssl and libcrypto 
+# on 64bits envirenments (it always find the 64bits version, regardless the architecture).
+# this is ugly, but the only way I fond :(
```
j4yk, line 26: Did you try https://cmake.org/cmake/help/v3.0/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html ?

estebanlm: well, this problem is to build a 32bits vm in a 64bits so I would need FIND_LIBRARY_USE_LIB32_PATHS... but anyway no, I didn't. I will try it, hopefully I can remove that ugliness :)

estebanlm: btw, this is fixed because now I'm also building openssl :S

estebanlm:
I will integrate this: errors in v3 are not related to my changes (Cog branch is failing with this problem too) and the pharo errors is some random write error I asked travis support because is something I cannot handle my self (if you have an idea, please tell me :)

PR #75, opened 2016-10-06
Simplifies the several shell scripts for building each Cog VM flavor …

santana:
…for Squeak into a single Makefile


krono:
While I like this, we wont go that way in the foreseeable future …

PR #78, opened 2016-10-09
OpenBSD support

santana:


fniephaus:
LGTM, what do you think @krono? Doesn't look like this is changing any of the original code paths, so I'd say this can be merged.

krono:
I'm ok in principle. But we have naming problems now. I think that's the reason @eliotmiranda wanted an `openbsd64x64`  (or so) dir. probably `unix64x64` would be better… I don't know, man :)

santana:
What naming problems? How is this change going to impact naming?

I don't think we need yet another directory. Maybe I can help if I can understand the problem.

fniephaus:
@krono I don't see why there's a need for an `openbsd64x64` or is it possible to emulate an OpenBSD build just like we are emulating ARM builds?
Other than that, there seems to be FreeBSD support already which is also part of `build.linux64x64`.

krono:
¯_(ツ)_/¯ Then the appropriate name would be unix….

Nevertheless, all changes to `Cross` and `unix` dirs are ok, and the changes to the `mvm` and such script are portable. OK from here

PR #79, opened 2016-10-09
Download PharoXXX.sources if its not found in the usual location.

bencoman:
If someone makes a fresh clone of the repository then the PharoXXX.sources file wont exist and a fresh-out-of-the-box Pharo build won't succeed.  So download it from its well known location.

### DIFF
--- a/build.linux32x86/editpharoinstall.sh
+++ b/build.linux32x86/editpharoinstall.sh
@@ -1,7 +1,8 @@
 #!/bin/bash -e
 # Edit the installed directory tree to rename squeak to pharo and install source
-INSTALLDIR="$1"
+OSVMROOTDIR=$(cd ../../..; pwd)
 
+INSTALLDIR="$1"
 shift
 cd $INSTALLDIR
 
@@ -11,8 +12,10 @@ if [ "$1" = -source ]; then
 else
 	SourceFile="PharoV50"
 fi
-SOURCE=../../sources/$SourceFile.sources
-test -f $SOURCE || SOURCE=../../../sources/$SourceFile.sources
+mkdir -p "$OSVMROOTDIR/sources"
+SOURCE=$OSVMROOTDIR/sources/$SourceFile.sources
+test -f $SOURCE || wget -O $SOURCE http://files.pharo.org/sources/$SourceFile
+
 if [ -f squeak ]; then
 	mv squeak pharo
 	sed -i 's/squeak/pharo/g' pharo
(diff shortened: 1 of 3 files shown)


build.linux32x86/editpharoinstall.sh
```diff
@@ -11,8 +11,11 @@ if [ "$1" = -source ]; then
 else
 	SourceFile="PharoV50"
 fi
-SOURCE=../../sources/$SourceFile.sources
-test -f $SOURCE || SOURCE=../../../sources/$SourceFile.sources
```
bencoman, line 15: btw, I'm curious why two folders are being tested for the sources file.  Why not standardise on one?  (In the root folder seems fine.)    

Also, a "sources" folder may confuse newcomers expecting C code. Maybe "imageSources" might be a better folder name, or maybe store them under "image/sources", or store sources in the image folder since SqueakV50.sources is already there ??

krono: As always, history. _sources_ is a well-known idea in the Smalltalk world, and "imageSource" actually does not cut it, as it is not the source from which images are built.

Btw: The Image looks for the sources file in several locations, also depending on platform.
Eg, for Mac it looks:
- in the current working directory
- in the directory of the `.image` file
- in the `Resources` folder of the App bundle
- in the directory the App bundle resides in
- (maybe something I forgot)

On Linux, is definitely also multiple locations…

bencoman: > The Image looks for the sources file in several locations

A bit different situation. This is not the Image looking for the sources file, but the build system finding the file to link next to the produced VM.

fniephaus: Maybe a copy and paste error? There used to be a [`sources` directory](http://www.squeakvm.org/svn/squeak/branches/Cog/sources/) in the root of the SVN repository, but we removed it when we moved to Git.

fniephaus:
Assigning this to @estebanlm since he is currently merging the Pharo VM back...

build.linux32x86/editpharoinstall.sh
bencoman: The root directory seems like the best place for the 'sources' folder. There are multiple build.xxx folders that use the files. That should be okay since the .sources files are gitignored.

eliotmiranda:
> On May 31, 2017, at 8:37 PM, Ben Coman <notifications@github.com> wrote:
> 
> @bencoman commented on this pull request.
> 
> In build.linux32x86/editpharoinstall.sh:
> 
> > @@ -11,8 +11,11 @@ if [ "$1" = -source ]; then
>  else
>  	SourceFile="PharoV50"
>  fi
> -SOURCE=../../sources/$SourceFile.sources
> -test -f $SOURCE || SOURCE=../../../sources/$SourceFile.sources
> The root directory seems like the best place for the 'sources' folder. There are multiple build.xxx folders that use the files. That should be okay since the .sources files are gitignored.
> 

+1

PR #81, opened 2016-10-16
replaces sprintf with snprintf

santana:
sprintf is usually misused; snprintf is safer in that regard.


bencoman (approved):
Good security initiative.  I've examined each of these and they look good to me.  In the outside chance a supported platform doesn't have snprintf, it should better to #define it once to sprintf, or add an open source version of snprintf (of which there are a few on github) to that platform.

nicolas-cellier-aka-nice:
Sounds good to me too, and no build is failing on tested platforms (the red crosses are unrelated).
It will be time to fix those eventual platforms not supporting snprintf when we encounter one (if ever).
If there is no other objection against this change, I'll merge this evening.

PR #88, opened 2016-11-01
configure.ac: Modernize scripts to allow autoreconf

zecke:
Allow to autoreconf with autoconf released in 2012. It might
be best to remove generated files to prevent bitrot again or
at least run it as part of the CI process.

Just tried to autoreconf but didn't run the script yet. Let's
see how it goes in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,8 @@ matrix:
 install:
 - ./.travis_install.sh
 script:
+- cd platforms/unix/config/ && make configure
+- cd ../../../
 - $CHROOT ./.travis_build.sh
 - ./.travis_test.sh
 after_success:
--- a/.travis_install.sh
+++ b/.travis_install.sh
@@ -69,7 +69,10 @@ sudo apt-get install -yq --no-install-suggests --no-install-recommends --force-y
      build-essential \
      python-dev \
      libffi-dev \
-     zlib1g-dev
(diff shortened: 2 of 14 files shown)


zecke:
Some of the link errors are from the fact that the MIDIPlugin now compiles against alsa (as configured) so..

```
/* this simply doesn't work */
#if defined(USE_MIDI_ALSA)
# include "sqUnixMIDIALSA.inc"
#else
```

the else part is not taken anymore. I don't think the buildsystem in this incarnation has a way to easily propagate link dependencies... but if MIDIPlugin is using asound.. and MIDIPlugin is built-in.. then squeak would need to link to asound...

a.) Link -lasound like -luuid -lpthread is passed
b.) Add a proper flag to enable/disable alsa in the MIDIPlugin and disable it to match status quo
c.) Improve buildsystem (and replicate automake+libtool)
d.) be happy that it "simply works" now?

eliotmiranda (approved):

eliotmiranda:
Holger, I hit confirm too soon.  Looks like the Cog version of the ALSA selection is correct.  can you resolve the conflict?

krono:
@zecke, any interest in seeing whether this works with current `Cog`?

krono:
This had been integrated in fec94cb812.
Close

PR #92, opened 2016-11-08
cache built thirdparty libraries 

estebanlm:
so they are built just once (otherwise they are built always and there is no point since they do not change).
also, introduce pharo.cog.spur for win32


estebanlm:
error is because of introduction of tests... nothing to do with PR

PR #101, opened 2016-11-24
add win32 build rules

estebanlm:


build.win32x86/common/Makefile
```diff
@@ -250,24 +260,25 @@ $(VMDEF) $(VMEXP) $(VMLIB): $(VMOBJ)
 .PHONY: $(OBJDIR)/%.lib $(OBJDIR)/%.dll
 
 # Check for Makefile in win32 plugins directory otherwise use default Makefile
-plugin-makefile = $(firstword $(realpath \
-					$(PLATDIR)/win32/plugins/$(*F)/Makefile.plugin \
-					../common/Makefile.plugin))
+plugin-makefile = $(realpath $(firstword $(wildcard $(WIN32PLUGINSDIR)/$(1)/Makefile.plugin ../common/Makefile.plugin)))
```
nicolas-cellier-aka-nice, line 263: 👍 should probably be back-ported to win32/win64 squeak/newspeak so as to enable build via MSYS

nicolas-cellier-aka-nice (approved):
Look OK from my eyes

build.win32x86/common/Makefile
estebanlm: is already there for all win32... it needs to be ported to win64 :)

estebanlm:
I will accept because the problem has nothing to do (and it will fixed with lib caching feature)

PR #105, opened 2016-12-05
fix PharoVM-win32 build + enhancements

estebanlm:
- added STACK_ALIGN_BYTES=16, ALLOCA_LIES_SO_USE_GETSP=0 flags to win32. Is a workaround, but it works for now.
- added a "third-party" directory to put "specs", to define common third-party library versions (this is first attempt to reduce the amount of duplicated code I have now).


estebanlm:
Error is with sista. Not related.

PR #109, opened 2016-12-21
Bug when calling mprotect and VirtualAlloc with the SpurMemoryManager

ronsaldo:
The endAddr parameter to sqMakeMemoryNotExecutableFromTo is exclusive. This means, that the size computation (endAddr - startAddr + 1) does not need the extra + 1.

When computing the size for mprotect and VirtualProtect with just endAddr - startAddr, the call does not fail anymore in Windows, Linux and OS X.

Please validate this with how are these functions used from the Slang side.

### DIFF
--- a/platforms/unix/vm/sqUnixSpurMemory.c
+++ b/platforms/unix/vm/sqUnixSpurMemory.c
@@ -1,19 +1,19 @@
 /* sqUnixSpurMemory.c -- dynamic memory management for Spur on unix & Mac OS X.
- * 
+ *
  * Author: eliot.miranda@gmail.com
- * 
+ *
  *   This file is part of Unix Squeak.
- * 
+ *
  *   Permission is hereby granted, free of charge, to any person obtaining a
  *   copy of this software and associated documentation files (the "Software"),
  *   to deal in the Software without restriction, including without limitation
  *   the rights to use, copy, modify, merge, publish, distribute, sublicense,
  *   and/or sell copies of the Software, and to permit persons to whom the
  *   Software is furnished to do so, subject to the following conditions:
- * 
+ *
  *   The above copyright notice and this permission notice shall be included in
  *   all copies or substantial portions of the Software.
- * 
+ *
  *   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


platforms/unix/vm/sqUnixSpurMemory.c
```diff
@@ -164,27 +164,27 @@ sqDeallocateMemorySegmentAtOfSize(void *addr, sqInt sz)
...
 {
-#	if 0
+	unsigned long size = endAddr - startAddr;
 	unsigned long firstPage = roundDownToPage(startAddr);
 	/* Arguably this is pointless since allocated memory always includes write
 	 * permission by default.  Annoyingly the mprotect call fails on both linux
 	 * and mac os x.  So make the whole thing a nop.
 	 */
 	if (mprotect((void *)firstPage,
-				 endAddr - firstPage + 1,
```
nicolas-cellier-aka-nice, line 184: But this is somehow different: the new size does not account for roundDownToPage

ronsaldo: You are right. The funny thing, is that the rounding was not being done at all at the Windows version. I wonder if the address were already always aligned to page boundaries. Anyway. I just tested by swapping the order of the lines, and it seems to be still working.

nicolas-cellier-aka-nice (approved):
It sounds good to me. My analysis of VMMaker senders also indicates that endAddress is exclusive.

A pity that we did not named the function sqMakeMemoryNotExecutableFrom:UpTo:

We might better add a comment to these functions

Also, it seems that an off by one error is made the other way around by senders of padIfPossibleWithStopsFrom:to: which is inclusive, but usage is (self padIfPossibleWithStopsFrom: start to: start + size) which is exclusive...

platforms/unix/vm/sqUnixSpurMemory.c
nicolas-cellier-aka-nice: man mprotect https://linux.die.net/man/2/mprotect : it requires an address aligned on page boundary
There is no such requirement for VirtualProtect https://msdn.microsoft.com/en-us/library/windows/desktop/aa366898(v=vs.85).aspx

I see from msdn page that we should FlushInstructionCache before executing any newly installed code and I fail to see any such call (grep -r FlushInstructionCache platforms). Unless the flush is performed by another mean?

PR #110, opened 2016-12-23
adds more dependencies to the CentOS list

santana:
and adds the -y flag so it can be copied and pasted and expected to work


eliotmiranda:
This is fine, but also needs to be made to the same sections in the other build.linux*/HowToBuild files.

fniephaus:
Could you update this PR please, @santana?

PR #120, opened 2017-02-21
Fix issue 119 jpeg plugin crashes in Win64

nicolas-cellier-aka-nice:
Solution is to align jump_buf on 16bytes boundary by using malloc/free pairs around setjmp/longjmp calls.
The struct error_mgr2 now contains a pointer to the jump_buf for this purpose.


platforms/Cross/plugins/JPEGReadWriter2Plugin/sqJPEGReadWriter2Plugin.c
```diff
@@ -54,9 +54,10 @@ primJPEGWriteImageonByteArrayformqualityprogressiveJPEGerrorMgrWriteScanlines(
 	error_ptr2 pjerr = (error_ptr2)jpegErrorMgr2Struct;
 
 	pcinfo->err = jpeg_std_error(&pjerr->pub);
+    pjerr->setjmp_buffer = (jmp_buf *) malloc(sizeof(jmp_buf));
```
nicolas-cellier-aka-nice: Hmm, I just noticed the difference of alignment, it seems like Xcode default set up was ruled by space lovers.
I'm committing with tabs.

nicolas-cellier-aka-nice:
The failures are completely unrelated to the fix

PR #121, opened 2017-02-21
Fixing and updating the Lowcode building scripts for Linux

ronsaldo:
In this pull request I am fixing and updating the Lowcode building scripts for linux.
I am also updating SDL2 to the latest version, because the old version is not building on Ubuntu 16.10.
I am regenerating spurlowcode64src because I was missing a method that produced undefined references in debug mode.


nicolas-cellier-aka-nice:
Hi Ronie, is this PR still relevant? If yes we can cherry pick the interesting bits.

ronsaldo:
I will move the interesting parts into a separate pull request. Closing this for now.

PR #129, opened 2017-04-04
fixes for Pharo

estebanlm:
- cairo for windows builds with correct link to freetype2
- libsdl2 for linux uses correct version (0.4.1)


nicolas-cellier-aka-nice:
The problem with this web interface is that we see all the changes of the merge from opensmalltalkvm->pharovm... We don't want to see them, they are already in opensmalltalk-vm tree obviously.
Exactly like recent changes from Tobias...
Ah, but I see that those merge are not marked as such in the network... So it's the workflow that is not really git friendly (et vice et versa).

This git/github interface is far far inferior to Monticello interface where only the diff to be applied are visible when we request a merge.

nicolas-cellier-aka-nice:
I've cherry picked all the relevant changes without conflict (these were false conflicts).
The remaining are merges from opensmalltalk-vm but not regular merges.
They are all marked as authored by Esteban which is a weird workflow (thus the conflicts).
They don't bring a single change. That's why I will close the pull request rather than merge, it's already merged.

estebanlm:
yeah, but now you cause a conflict in my sources :)
(yes, I agree there is something weird in the diff tool of github... it shoudln't show differences in the merged files... they are already there).

nicolas-cellier-aka-nice:
I Esteban, just merge opensmalltalk-vm/cog using git, there should not be any conflict.

PR #130, opened 2017-04-06
pharo to deploy from opensmalltalk-vm

estebanlm:
again, is showing me a lot of changed sources which actually are not part of the change :(


estebanlm:
ah no, now is ok :) 
(there are somethings very weird with this PRs :P )

estebanlm:
ok, let's see how this works :)

PR #134, opened 2017-04-25
added uuid-dev and other packages essential on Ubuntu

kksubbu:
Added uuid-dev and other packages to let configure step proceed without failure on Ubuntu 64b. Without uuid-dev, configure will report an error about missing -luuid and stop.


kksubbu:
Will submit pulls from a separate topic branch

PR #136, opened 2017-05-07
command line arguments use single dash prefix but also accept doubledash

kksubbu:
Unix conventions use -- for word options and - for single letter options. But this applies only to Unix commands and not to virtual machines like qemu or squeak.

PharoVM compile time macro replaced - with --. These modifications use a single dash in all help messages while recovering gracefully if two dashes are used for prefixes accomodating both classic squeak and the newer pharo usage.

Using -help or -version no longer exits with non-zero codes as they are valid usages.

### DIFF
--- a/platforms/unix/vm/sqUnixMain.c
+++ b/platforms/unix/vm/sqUnixMain.c
@@ -1378,7 +1378,7 @@ static void vm_parseEnvironment(void)
 }
 
 
-static void usage(int exitValue);
+static void usage();
 static void versionInfo(void);
 
 
@@ -1464,7 +1464,7 @@ static int vm_parseArgument(int argc, char **argv)
 
   /* vm arguments */
 
-  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage(0);         /*NOTREACHED*/}
+  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage();         	exit(0); }
   else if (!strcmp(argv[0], VMOPTION("version")))	{ versionInfo();	return 0; }
   else if (!strcmp(argv[0], VMOPTION("noevents")))	{ noEvents	= 1;	return 1; }
   else if (!strcmp(argv[0], VMOPTION("nomixer")))	{ noSoundMixer	= 1;	return 1; }
@@ -1689,7 +1689,7 @@ SqModuleDefine(vm, Module);
 /*** options processing ***/
 
 
-static void usage(int exitValue)
+static void usage()
 {
   struct SqModule *m= 0;
   printf("Usage: %s [<option>...] [<imageName> [<argument>...]]\n", argVec[0]);
@@ -1719,7 +1719,6 @@ static void usage(int exitValue)
   printf("\nAvailable drivers:\n");
   for (m= modules;  m->next;  m= m->next)
     printf("  %s\n", m->name);
-  exit(exitValue);
 }
 
 
@@ -1817,8 +1816,8 @@ static void parseArguments(int argc, char **argv)
       if (n == 0)			/* option not recognised */
 	{
 	  fprintf(stderr, "unknown option: %s\n", argv[0]);
-	  usage(1);
-	  /*NOTREACHED*/
+	  usage();
+	  exit(1);
 	}
       while (n--)
 	saveArg();
--- a/platforms/unix/vm/sqUnixMain.c
+++ b/platforms/unix/vm/sqUnixMain.c
@@ -1310,9 +1310,6 @@ static void loadModules(void)
   checkModuleVersion(soundModule,   SqSoundVersion,   snd->version);
 }
 
-/* built-in main vm module */
-
-
 static long
 strtobkm(const char *str)
 {
@@ -1345,6 +1342,7 @@ static int jitArgs(char *str)
 }
 #endif /* !STACKVM && !COGVM */
 
+/* ----------------- built-in main vm module */
 
 # include <locale.h>
 static void vm_parseEnvironment(void)
@@ -1380,7 +1378,7 @@ static void vm_parseEnvironment(void)
 }
 
 
-static void usage(void);
+static void usage(int exitValue);
 static void versionInfo(void);
 
 
@@ -1398,7 +1396,11 @@ static int parseModuleArgument(int argc, char **argv, struct SqModule **addr, ch
 
 static int vm_parseArgument(int argc, char **argv)
 {
-  /* deal with arguments that implicitly load modules */
+  // parse arguments for main vm module including those that
+  // implicitly load modules.
+
+  if (argv[0][0] == '-' && argv[0][1] == '-')
+	argv[0]++;	/* skip one dash in double dash options */
 
   if (!strncmp(argv[0], "-psn_", 5))
     {
@@ -1418,13 +1420,8 @@ static int vm_parseArgument(int argc, char **argv)
       return 1;
     }
 
-  /* legacy compatibility */		/*** XXX to be removed at some time ***/
 
-#ifdef PharoVM
-# define VMOPTION(arg) "--"arg
-#else
 # define VMOPTION(arg) "-"arg
-#endif
 
 # define moduleArg(arg, type, name)						\
     if (!strcmp(argv[0], VMOPTION(arg)))							\
@@ -1467,7 +1464,8 @@ static int vm_parseArgument(int argc, char **argv)
 
   /* vm arguments */
 
-  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage();		return 1; }
+  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage(0);         /*NOTREACHED*/}
+  else if (!strcmp(argv[0], VMOPTION("version")))	{ versionInfo();	return 0; }
   else if (!strcmp(argv[0], VMOPTION("noevents")))	{ noEvents	= 1;	return 1; }
   else if (!strcmp(argv[0], VMOPTION("nomixer")))	{ noSoundMixer	= 1;	return 1; }
   else if (!strcmp(argv[0], VMOPTION("notimer")))	{ useItimer	= 0;	return 1; }
@@ -1481,7 +1479,6 @@ static int vm_parseArgument(int argc, char **argv)
   else if (!strcmp(argv[0], VMOPTION("nojit")))		{ useJit	= 0;	return 1; }
   else if (!strcmp(argv[0], VMOPTION("spy")))		{ withSpy	= 1;	return 1; }
 #endif /* !STACKVM && !COGVM */
-  else if (!strcmp(argv[0], VMOPTION("version")))	{ versionInfo();	return 1; }
   else if (!strcmp(argv[0], VMOPTION("single")))	{ runAsSingleInstance=1; return 1; }
   /* option requires an argument */
   else if (argc > 1)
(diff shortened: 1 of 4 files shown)


platforms/unix/vm/sqUnixMain.c
```diff
@@ -1467,7 +1464,8 @@ static int vm_parseArgument(int argc, char **argv)
 
   /* vm arguments */
 
-  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage();		return 1; }
+  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage(0);         /*NOTREACHED*/}
```
bencoman: Would it be better to shift the exit() out of usage(), so that real code affection execution flow is used here rather than a comment.   i.e. here...  { usage();   exit(0) }  ??

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1721,7 +1719,7 @@ static void usage(void)
   printf("\nAvailable drivers:\n");
   for (m= modules;  m->next;  m= m->next)
     printf("  %s\n", m->name);
-  exit(1);
+  exit(exitValue);
```
bencoman: Refactor the exit() out to the caller ??

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1819,7 +1817,8 @@ static void parseArguments(int argc, char **argv)
       if (n == 0)			/* option not recognised */
 	{
 	  fprintf(stderr, "unknown option: %s\n", argv[0]);
-	  usage();
+	  usage(1);
+	  /*NOTREACHED*/
```
bencoman: Same as above, refactor exit out of usage() ?   i.e.  here...
     usage(); 
     exit(1)

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1692,11 +1689,12 @@ SqModuleDefine(vm, Module);
 /*** options processing ***/
 
 
-static void usage(void)
+static void usage(int exitValue)
 {
   struct SqModule *m= 0;
   printf("Usage: %s [<option>...] [<imageName> [<argument>...]]\n", argVec[0]);
   printf("       %s [<option>...] -- [<argument>...]\n", argVec[0]);
+  printf("options begin with single -, but -- prefix is silently accepted\n");
```
bencoman, line 1694: The problem is that single-hyphen default is then what is advised by --help.  Why isn't it preferable to conform to Unix conventions [1] "The GNU double-hyphen option leader was chosen so that traditional single-letter options and GNU-style keyword options could be unambiguously mixed on the same command line. "  
[1] http://www.catb.org/~esr/writings/taoup/html/ch10s05.html

The single-hyphen could be silently accepted for backward compatibility. Maybe better would be a deprecated message, but perhaps that could adversely affect existing scripts(?).

How do Squeakers feel about that?

bencoman (changes requested):
Can we have some discussion on [vm-dev] about the flexibility of Squeakers to use double-hyphen options as default?

eliotmiranda:
IIABDFI
Esteban rewrote the code that use -- for Pharo, preserving - for Squeak.  What's the problem??

_,,,^..^,,,_ (phone)

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1467,7 +1464,8 @@ static int vm_parseArgument(int argc, char **argv)
 
   /* vm arguments */
 
-  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage();		return 1; }
+  if      (!strcmp(argv[0], VMOPTION("help")))		{ usage(0);         /*NOTREACHED*/}
+  else if (!strcmp(argv[0], VMOPTION("version")))	{ versionInfo();	return 0; }
```
bencoman: [Edit] Duh, its not added but relocated.  
But then, why change to return 0  /* option not recognised */   ??
The existing seems to work okay...
Pharo 5 $     ./pharo --version > /dev/null ; echo $?
0
Pharo6-stable-threaded-20170503 $    ./pharo --version > /dev/null ; echo $?
0
Squeak5.1-16549-32bit-20160817 $    ./squeak.sh -version > /dev/null ; echo $?
0

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1418,13 +1420,8 @@ static int vm_parseArgument(int argc, char **argv)
       return 1;
     }
 
-  /* legacy compatibility */		/*** XXX to be removed at some time ***/
 
-#ifdef PharoVM
-# define VMOPTION(arg) "--"arg
-#else
 # define VMOPTION(arg) "-"arg
-#endif
```
bencoman, line 1422: At least for now, please revert this #ifdef per Eliot's feedback "IIABDFI" [1].
This will facilitate the rest of PR proceeding.  (also, its mixing concerns.) 

btw, What was the reason to have both work single and double dashes working in the one VM?

[1] http://forum.world.st/Re-OpenSmalltalk-opensmalltalk-vm-command-line-arguments-use-single-dash-prefix-but-also-accept-doub-td4945805.html

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1398,7 +1396,11 @@ static int parseModuleArgument(int argc, char **argv, struct SqModule **addr, ch
 
 static int vm_parseArgument(int argc, char **argv)
 {
-  /* deal with arguments that implicitly load modules */
+  // parse arguments for main vm module including those that
+  // implicitly load modules.
+
+  if (argv[0][0] == '-' && argv[0][1] == '-')
+	argv[0]++;	/* skip one dash in double dash options */
```
bencoman: The problem with this is that the help displays single dashes for Pharo.
This is not necessarily desired and would require further discussion on [vm-dev] mail list. 
Please revert so the --help part of this PR can proceed.

platforms/unix/vm/sqUnixMain.c
bencoman: I see current usage is...
Pharo5$    ./pharo --help > /dev/null ; echo $?
0
Pharo6-stable-threaded-20170503$    ./pharo --help > /dev/null ; echo $?
1
Squeak5.1-16549-32bit-20160817$    ./squeak.sh -help > /dev/null ; echo $?
1

Eliot has approved Squeak exit of 0 for -help.

platforms/unix/vm/sqUnixMain.c
bencoman: Based on limited information on the need for this, Eliot's reply "leave it as is" [1] .
Please revert and if there is pressing need, raise discussion on [vm-dev] list.  

[1] http://forum.world.st/Re-OpenSmalltalk-opensmalltalk-vm-command-line-arguments-use-single-dash-prefix-but-also-accept-doub-td4945805.html

platforms/unix/vm/sqUnixMain.c
kksubbu: Good idea. Usually, usage is a fatal error and is not expected to return. But in this case, moving exit out improves readability. Thanks.

platforms/unix/vm/sqUnixMain.c
kksubbu: See the explanation in the first comment. Unix apps allow multiple single letter options to merge into one word (e.g. grep -lir for grep -l -i -r), so GNU introduced an extra dash for word options. Multi-platform virtual machines like qemu-* or squeak don't allow collapse of single letter options, so an extra dash is superfluous. It does eat into max line length without adding any value.

I have no stakes either way. I just want to fix it once, skip bike-shedding and move onto other important stuff.

platforms/unix/vm/sqUnixMain.c
kksubbu: I just relocated it to keep it closer to help, so it is easy to see that help/version are valid options that exit with 0.

kksubbu:
A quick check for the patch:
```
$ cd oscogvm/build.linux64x64/squeak.cog.spur/build
$ for i in help version; do ./squeak -$i >/dev/null && echo pass; ./squeak --$i >/dev/null && echo pass; done
pass
pass
pass
pass
$ for i in help version; do ./squeak ---$i >/dev/null || echo pass; done
unknown option: --help
pass
unknown option: --version
pass
```

bencoman:
But... $ pharo --help 
shows...
     -encoding <enc>       set the internal character encoding (default: MacRoman)
instead of...
     --encoding <enc>       set the internal character encoding (default: MacRoman)
I don't know whether that is critical.  Its Esteban's call.

bencoman:
Also, I'm (only mildly) curious why the original... 
   { versionInfo();	return 1; }
was changed to 
   { versionInfo();	return 0; }
when at the end of vm_parseArgument() is has
  return 0;	/* option not recognised */

Anyway it seems to have little impact.

kksubbu:
On Tuesday 09 May 2017 06:04 PM, Ben Coman wrote:
> But... $ pharo --help
> shows...
> -encoding set the internal character encoding (default: MacRoman)
> instead of...
> --encoding set the internal character encoding (default: MacRoman)
> I don't know whether that is critical. Its Esteban's call.

This is intentional. -encoding is the option, but --encoding is also 
accepted as explained near the beginning of the help text.

Regards .. Subbu

kksubbu:
On Tuesday 09 May 2017 06:22 PM, Ben Coman wrote:
> Also, I'm (only mildly) curious why the original...
> { versionInfo(); return 1; }
> was changed to
> { versionInfo(); return 0; }
> when at the end of vm_parseArgument() is has
> return 0; /* option not recognised */

this is because it could be valid option for another module/plugin. See
----
  modulesDo (m)
        if ((n= m->parseArgument(argc, argv)))
           break;
----
options which are not picked up by any module are caught just below this 
code. Something like

   +<plugin>=opt1,opt2:value,...

would be more robust in avoid name clashes across modules, but the code 
will get even more complex.

Regards .. Subbu

kksubbu:
Fixed vm extensions to handle -/-- prefixes for options gracefully. Here is the smoke test script:
```
#!/bin/sh
for i in help version; do
	./squeak -$i >/dev/null && echo pass -$i
	./squeak --$i >/dev/null && echo pass --$i
	./squeak ---$i >/dev/null || echo pass ---$i
done &&
# test options for vm X11 and sound
for i in headless swapbtn jit nomixer; do
	./squeak --$i --help >/dev/null && echo pass --$i
	./squeak ---$i --help >/dev/null || echo pass ---$i
	./squeak ---$i >/dev/null || echo pass ---$i
done &&
# test options with arg
for i in memory title; do
	./squeak --$i 1024 --help >/dev/null && echo pass --$i
	./squeak ---$i 1024 --help >/dev/null || echo pass ---$i
	./squeak ---$i 1024 >/dev/null || echo pass ---$i
done && echo pass all
```

platforms/unix/vm/sqUnixMain.c
```diff
@@ -1481,7 +1476,6 @@ static int vm_parseArgument(int argc, char **argv)
   else if (!strcmp(argv[0], VMOPTION("nojit")))		{ useJit	= 0;	return 1; }
   else if (!strcmp(argv[0], VMOPTION("spy")))		{ withSpy	= 1;	return 1; }
 #endif /* !STACKVM && !COGVM */
-  else if (!strcmp(argv[0], VMOPTION("version")))	{ versionInfo();	return 1; }
```
bencoman, line 1484: (deleted) Sorry I was late to understand my question here was already addressed.

bencoman:
I'm not sure my bike-shedding helped anything. Anyway, while this is waiting for authoritative go/nogo from a committer, I've done a another review just now and I'm satisfied with all the answers to my queries.

bencoman (approved):

platforms/unix/vm/sqUnixMain.c
kksubbu: I removed the #ifdef because dash prefix is not a compile time option. Modules have their own parser for arguments :-(. While #ifdef seems like a simple fix for the core vm, it complicates argument parsing in modules, particularly those which are shared between Squeak and Pharo :-(. It also affects hard-coded messages like the one in imageNotFound().

PR #137, opened 2017-05-07
[skip ci] uuid-dev is needed on Ubuntu for configure script.

kksubbu:
Updated build instructions to add uuid-dev and GL libraries for configure script to run on Ubuntu. Without uuid-dev configure would fail with an error message about missing uuid library.

### DIFF
--- a/build.linux64x64/HowToBuild
+++ b/build.linux64x64/HowToBuild
@@ -248,7 +248,6 @@ and then retry.
 
 
 Ubuntu
-	sudo apt-get install libcairo2-dev   
-	sudo apt-get install libpango1.0-dev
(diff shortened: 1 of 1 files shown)


eliotmiranda (approved):
Hi Subbu; this should surely be in all the linux HowToBuilds, no?

kksubbu:
On Tuesday 09 May 2017 12:04 AM, Eliot Miranda wrote:
> Hi Subbu; this should surely be in all the linux HowToBuilds, no?

Yes. I just haven't tested it on all variants.

IMHO, it is better to have one build folder for all linux builds and 
handle variants like insruction set architecture, wordsize etc. within 
that build. Linux has a much large number of variants compared to other 
platforms, but it is easy to detect them on the fly (e.g. package name 
varying by distro) through scripts/* and makefiles.

Regards .. Subbu

PR #142, opened 2017-06-21
updates libgit2 for pharo-vm and some other plugins

estebanlm:
also it adds files for compiling on freebsd, etc.


estebanlm:
merged because is ok :)

PR #143, opened 2017-07-03
Add checkoutVMbyDate

akgrant43:
checkoutVMbyDate is a convenience script to check out the commit of a
particular VM.


kksubbu:
Alistair,

You could use awk to simplify extraction of date string:
```
vs1=$($pharoexe --version | awk -F\$ '/Date/ {print substr($2,8,24)}'
```
For example, 
```
$ ./pharo Pharo --version | awk -F\$ '/Date/ {print substr($2,8,24)}'
Tue May 30 19:41:27 2017
```

PR #144, opened 2017-07-18
sqUnixHeartbeat.c: Remove warning about thread priorities

lukego:
JFYI: This is a patch that I applied on my own application-specific branch.

----

The default behavior is to print a loud warning about thread
priorities and ask the user to update the Linux kernel configuration
to allow Squeak a higher-priority thread.

    pthread_setschedparam failed: Operation not permitted
    This VM uses a separate heartbeat thread to update its internal clock
    and handle events.  For best operation, this thread should run at a
    higher priority, however the VM was unable to change the priority.  The
    effect is that heavily loaded systems may experience some latency
    issues.  If this occurs, please create the appropriate configuration
    file in /etc/security/limits.d/ as shown below:

    cat <<END | sudo tee /etc/security/limits.d/pharo.conf
    *      hard    rtprio  2
    *      soft    rtprio  2
    END

    and report to the pharo mailing list whether this improves behaviour.

    You will need to log out and log back in for the limits to take effect.
    For more information please see
    https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/tag/r3732#linux

I am removing this message for a combination of reasons:

- Is there really a problem? How to demonstrate it?
- Is this really a solution? How to demonstrate it?
- Is there potential for unintended system-wide consequences?
- Could we _lower_ the prio of the main thread instead?
- No option to suppress this message if you don't want to see it?

Generally I am skeptical that Linux needs any scheduler tweaks for
~millisecond response time (that's not very tight) and if the system
really _is_ overloaded then I don't think it's safe to assume the user
wants Pharo to clobber all the other work.

The printed instructions are also distro-specific and won't work on
NixOS, which makes it all the more bothersome to see that message
every single time Pharo is started, and in all log files, etc...

### DIFF
--- a/platforms/unix/vm/sqUnixHeartbeat.c
+++ b/platforms/unix/vm/sqUnixHeartbeat.c
@@ -320,25 +320,6 @@ beatStateMachine(void *careLess)
 #else
 # define VMNAME "squeak"
 #endif
-        fprintf(stderr, "This VM uses a separate heartbeat thread to update its internal clock\n");
-        fprintf(stderr, "and handle events.  For best operation, this thread should run at a\n");
-        fprintf(stderr, "higher priority, however the VM was unable to change the priority.  The\n");
-        fprintf(stderr, "effect is that heavily loaded systems may experience some latency\n");
-        fprintf(stderr, "issues.  If this occurs, please create the appropriate configuration\n");
-        fprintf(stderr, "file in /etc/security/limits.d/ as shown below:\n\n");
-        fprintf(stderr, "cat <<END | sudo tee /etc/security/limits.d/%s.conf\n", VMNAME);
-        fprintf(stderr, "*      hard    rtprio  2\n");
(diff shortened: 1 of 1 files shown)


jecisc:
What is the status of the PR? It would be good to have less warning. It create a lot of noise in CI logs.

fniephaus:
> Is there really a problem? How to demonstrate it?

According to https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/275#issuecomment-411558942 and https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/275#issuecomment-411569935, there is.

I'm afraid we can't just remove that message. You should, however, be able to suppress the warning by redirecting `stderr` to `/dev/null/` (via `2>/dev/null`). I would not be opposed to adding a flag for this as well (e.g. `-suppress-rtprio-warning`).

Closing this in favor of #275.

PR #147, opened 2017-08-25
Issue 20294 Add FileAttributesPlugin to the linux 32 bit VM

akgrant43:
Add the windows support code for the FileAttributesPlugin:

platforms/win32/plugins/FileAttributesPlugin/sqWin32File.h


eliotmiranda:
Hi Alistair,

I think better than duplicating sqWin32File.h is to add a Makefile.plugin to add FilePlugin as an include path.  It would look like this:

platforms/win32/plugins/FileAttributesPlugin/Makefile.plugin:
INCLUDES += -I$(WIN32PLUGINSDIR)/FilePlugin

See for example platforms/win32/plugins/Win32OSProcessPlugin/Makefile.plugin

akgrant43:
Hi Eliot,

Thanks!  That's a much better solution.

I've updated the branch to add the makefile and removed the copy of sqWin32File.h.  The PR should contain 2 commits, with one file changed (the addition of Makefile.plugin).

Thanks again,
Alistair

nicolas-cellier-aka-nice:
Just note that failures are unrelated, so this should not prevent acceptance of this PR
For Pharo win64 it's the obsolete libpng link problem
For non windows spur Squeak, it's the testBecomeHash that fails.

akgrant43:
Nicolas, thanks for your review and approval.

PR #148, opened 2017-08-27
20341-libpng-source-is-no-longer-available

akgrant43:
The source at ftp.simplesystems.org seems to have gone away.

Based on information at http://www.libpng.org/pub/png/libpng.html
update the download url to point to sourceforge and update to 1.6.30.

### DIFF
--- a/third-party/libpng.spec.win
+++ b/third-party/libpng.spec.win
@@ -2,9 +2,9 @@
 # because 1.6.28 may fail on macOS (1.6.26 did).
 # this cannot be good and eventually it will be a source of problems... so we need to reunite 
 # versions as soon as possible.
-libpng_spec_download_url:=ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.29.tar.gz
-libpng_spec_archive_name:=libpng-1.6.29.tar.gz
-libpng_spec_unpack_dir_name:=libpng-1.6.29
+libpng_spec_download_url:=http://prdownloads.sourceforge.net/libpng/libpng-1.6.30.tar.gz
+libpng_spec_archive_name:=libpng-1.6.30.tar.gz
+libpng_spec_unpack_dir_name:=libpng-1.6.30
 libpng_spec_product_name_macOS:=libpng16.16.dylib
 libpng_spec_product_name_linux:=
 # Yes, 16-16 for some weird reason
--- a/third-party/libpng.spec.win
+++ b/third-party/libpng.spec.win
@@ -2,9 +2,9 @@
 # because 1.6.28 may fail on macOS (1.6.26 did).
 # this cannot be good and eventually it will be a source of problems... so we need to reunite 
 # versions as soon as possible.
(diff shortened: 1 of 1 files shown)


nicolas-cellier-aka-nice:
As usual, failures are unrelated...
But why change the ftp address to http?
And why 1.6.30 and not 1.6.32?
It seems there are later vulnerability fixes.

akgrant43:
Hi Nicolas,

> But why change the ftp address to http?

Only because that's the address I found on what I believe is the
official home page (happy to be corrected).


> And why 1.6.30 and not 1.6.32?

Sigh...  Yesterday .30 *was* the latest version.

Given the comment about .31 I assume they reverted back to .30 as soon
as they found the bug, and have now resolved it in .32.


> It seems there are later vulnerability fixes.

I'll update the branch and re-test (hopefully tonight).

Cheers,
Alistair

nicolas-cellier-aka-nice:
Hi Alistair,
I suggest you let the ftp address unchanged (it still is valid and linked from there front page), switch to 1.6.32, and I'll accept the new PR immediately 👍  (hoping we'll be faster than libpng team)

akgrant43:
Hi Nicolas,

> I suggest you let the ftp address unchanged (it still is valid and linked from there front page)

I can't find any reference to simplesystems, i.e.:

ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.x.tar.gz

from: http://www.libpng.org/pub/png/libpng.html (which is the official home page?)

The reason I changed to sourceforge was that what was the then current version (.30) wasn't available at simplesystems.  My guess is that it was still hosting .31 (the broken version).

Sourceforge also seems to be much faster than simplesystems for me, 0.5s vs 91s

> switch to 1.6.32, and I'll accept the new PR immediately :+1: (hoping we'll be faster than libpng team)

Sure, I'll change it back to simplesystems, 0.32, do a test build and let you know.

Cheers,
Alistair

nicolas-cellier-aka-nice:
Oh, I see, they switched, as explained at https://libpng.sourceforge.io/

> The previous ftp site is still being maintained at ftp://ftp.simplesystems.org/pub/png, but the OSUOSL site is much faster and has a more complete archive of old versions. 

I just checked that 1.6.32 was available at ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/
I cannot list the new ftp site thru fireftp client, while it works for the old (???)
So for the moment I suggest keeping the old ftp site as long as it is maintained...

akgrant43:
Hi Nicolas,

Done...

You should see 2 commits with 1 file changed (just 29 -> 32).

Just FYI, using Chrome I can get a directory listing from both:

ftp://ftp.simplesystems.org/pub/png/src/libpng16/
ftp://ftp-osl.osuosl.org/pub/libpng/src/libpng16/

Cheers,
Alistair

nicolas-cellier-aka-nice:
OK, that's just fireftp which hangs, I can see the ftp contents thru MacOSX Finder

akgrant43:
Thanks, Nicolas!

PR #155, opened 2017-10-07
Added X11 support for horizontal scroll wheel.

michael-roe:
Added X11 support for mice with a horizontal scroll wheel, such as Apple mice.

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -3554,6 +3554,8 @@ static const char *
 nameForKeyboardEvent(XEvent *evt) { return nameForKeycode(evt->xkey.keycode); }
 #endif /* DEBUG_EVENTS */
 
+static int mouseWheel2Squeak[4] = {30, 31, 28, 29};
+
 static void handleEvent(XEvent *evt)
 {
 #if DEBUG_EVENTS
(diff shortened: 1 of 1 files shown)


michael-roe:
In the Pharo bug tracker, issue #4795 "Horizontal wheel events" explains why this fix will be problematic: control+left/right arrow are used for other purposes (moving by a word in the text editor), so we can't use this key combination to represent a scroll wheel event without breaking its other use as a shortcut in the text editor.

In the Mac OS virtual machine, fakeMouseWheelKeyboardEvents() in sqMacUIEvents.c has an #ifdef PharoVM to use control+option+command+shift+cursor key for mouse wheel events, avoiding the clash with the text editor's keyboard shortcut. We could do this in the Unix virtual machine as well.

krono:
@eliotmiranda, any thoughts here? Otherwise, I'd just merge and see

eliotmiranda:
No thoughts.  I’m not doing much work on Linux these days.  So I agree.  Let’s merge and see what happens.

eliotmiranda:
and on iOS (the Cocoa Mac OS/iPhone platforms) the same approach is taken in fakeMouseWheelKeyboardEventsKeyCode:windowIndex:

    evt.modifiers = (CtrlKeyBit | OptionKeyBit | CommandKeyBit | ShiftKeyBit);

So I shall modify the X11 code to do the same.  And I shall comment the code.  These things NEED TO BE COMMENTED :-)

PR #156, opened 2017-10-09
openssl: Upgrade to minor release "l" to address multiple issues

zecke:
According to https://www.openssl.org/news/vulnerabilities.html#y2017
this should fix:

CVE-2017-3731, CVE-2017-3732, CVE-2016-7055.

The download should have the sha256sum of:
ce07195b659e75f4e1db43552860070061f156a98bb37b672b101ba6e3ddf30c

### DIFF
--- a/third-party/openssl.spec
+++ b/third-party/openssl.spec
@@ -1,6 +1,6 @@
-openssl_spec_download_url:=https://www.openssl.org/source/openssl-1.0.2j.tar.gz
-openssl_spec_archive_name:=openssl-1.0.2j.tar.gz
-openssl_spec_unpack_dir_name:=openssl-1.0.2j
+openssl_spec_download_url:=https://www.openssl.org/source/openssl-1.0.2l.tar.gz
+openssl_spec_archive_name:=openssl-1.0.2l.tar.gz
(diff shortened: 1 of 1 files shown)


nicolas-cellier-aka-nice:
+1 for this change but is it sustainable?
I understand that declaring a specific version is a good recipe for having reproducible artefacts.
But we are then taking responsibilities wrt security and somehow duplicating the work made in OS distributions. Can we really assume these responsibilities, or should we employ our scarce resources to more Smalltalk-specific tasks?

zecke:
Well my personal opinion doesn't really matter here. As OpenSSL is not a system library on Windows OpenSSL we" have assumed this responsibility. Luckily for freetype2, libpng, and OpenSSL (which are currently vulnerable) the risk of upgrade is low (API stability, minor versions are maintained)

nicolas-cellier-aka-nice:
Funnily enough, it seems that the build failed not because of your PR, but because wget libpng failed (they have a strict policy: remove their tarball as soon as a security issue is fixed). I still think that we can't run after such train...

Note that since we use the cygwin distribution for mingw target, (the mingw version of) openssl is built and updated by cygwin folks... So even if it is not purely OS provided, it's cared of by maintainers one layer above.

PR #161, opened 2017-10-27
fix the preparation of windows paths for dir_EntryLookup

estebanlm:
if path+name was bigger than 260 but name was smaller, then the conversion was not being applied.

This breaks the function when trying to check for files that actually valid.

### DIFF
--- a/platforms/win32/vm/sqWin32Directory.c
+++ b/platforms/win32/vm/sqWin32Directory.c
@@ -381,6 +381,7 @@ sqInt dir_EntryLookup(char *pathString, sqInt pathLength, char* nameString, sqIn
   FILETIME fileTime;
   SYSTEMTIME sysTime;
   sqInt sz, fsz;
+  char *fullName;
 #ifdef PharoVM
   HANDLE findHandle;
   sqInt i;
@@ -436,16 +437,13 @@ sqInt dir_EntryLookup(char *pathString, sqInt pathLength, char* nameString, sqIn
     return BAD_PATH;
   }
 
-  /* Ensure trailing delimiter and add filename. */
-  if(win32Path[win32PathLength-1] != L'\\') win32PathLength++;
-  fsz = MultiByteToWideChar(CP_UTF8, 0, nameString, nameStringLength, NULL, 0);
-  sz = win32PathLength;
-  win32PathLength += fsz;
-  if(win32PathLength >= 32767) FAIL();
-  REALLOC_WIN32_PATH(win32Path, win32PathLength);
-  win32Path[win32PathLength-fsz-1] = L'\\';
-  MultiByteToWideChar(CP_UTF8, 0, nameString, nameStringLength, win32Path+sz, fsz); 
-  win32Path[win32PathLength] = 0; // Not needed. See REALLOC_WIN32_PATH.
+  fullName = alloca(pathLength + nameStringLength + 1);
+  strcpy(fullName, pathString);
+  if(pathString[pathLength] != '\\') {
+	  strcat(fullName, "\\");
+  }
+  strcat(fullName, nameString);
+  ALLOC_WIN32_PATH(win32Path, fullName, pathLength + nameStringLength + 1);
 
   if(!GetFileAttributesExW(win32Path, 0, &winAttrs)) {
 #ifdef PharoVM


jecisc:
This bug this PR correct is really bothering for industrial applications to deploy on Windows. Thank you!

platforms/win32/vm/sqWin32Directory.c
```diff
@@ -436,16 +437,13 @@ sqInt dir_EntryLookup(char *pathString, sqInt pathLength, char* nameString, sqIn
...
-  if(win32Path[win32PathLength-1] != L'\\') win32PathLength++;
-  fsz = MultiByteToWideChar(CP_UTF8, 0, nameString, nameStringLength, NULL, 0);
-  sz = win32PathLength;
-  win32PathLength += fsz;
-  if(win32PathLength >= 32767) FAIL();
-  REALLOC_WIN32_PATH(win32Path, win32PathLength);
-  win32Path[win32PathLength-fsz-1] = L'\\';
-  MultiByteToWideChar(CP_UTF8, 0, nameString, nameStringLength, win32Path+sz, fsz); 
-  win32Path[win32PathLength] = 0; // Not needed. See REALLOC_WIN32_PATH.
+  fullName = alloca(pathLength + nameStringLength + 1);
```
nicolas-cellier-aka-nice, line 440: `+1` here is for trailing null character (usage of strcpy oblige), isn't it?
it does not account for potentially missing \ that is added below, does it?

platforms/win32/vm/sqWin32Directory.c
```diff
@@ -436,16 +437,13 @@ sqInt dir_EntryLookup(char *pathString, sqInt pathLength, char* nameString, sqIn
...
+  strcpy(fullName, pathString);
```
nicolas-cellier-aka-nice, line 441: But is pathString null terminated at this stage, or rather a Smalltalk (utf8-encoded?) ByteString?

nicolas-cellier-aka-nice:
OK, it seems that you are after the right bug.
If I rephrase: `ALLOC_WIN32_PATH` macro does handle path longer than `MAX_PATH` with `\\?\` prefix
When path is shorter, but path+name is longer than `MAX_PATH`, we go thru  `REALLOC_WIN32_PATH` which does nothing for long path...

However, I don't think that the fix is correct...
IMO, the fix could be done using strncpy with prescribed Smalltalk string length
and we don't need a terminating NULL character since ALLOC_WIN32_PATH does not require one

nicolas-cellier-aka-nice:
I have pushed a concurrent solution in the FIX_WIN32_LONG_FILENAME branch.
It rather fix the REALLOC_WIN32_PATH macro.
It may seem more complex than strncpy solution though.

estebanlm:
I really don't care as long as the problem is solve, but... why to use macros when we could use functions? 
They are a lot easier to understand/debug.
What's the gain on rely on macros?

nicolas-cellier-aka-nice:
I think that using a macro was essentially for using alloca instead of malloc.
Using alloca in a function that would then return a dangling pointer would not be a good idea, would it?
The macro can also directly invoke `FAIL()`, while I'm not so sure that a function could. We would have to write more code for checking invalid output after calling the function, so the macro somehow factor that code.

I have chosen to change the `REALLOC_WIN32_PATH` macro for 2 reasons:
- it is used twice (in the directory lookup too)
- I overlooked all the `#if !defined(WIN_CE)` part, and thought it was safer to make a more local change for not breaking anything.

Now having two similar macros doing essentially the same thing smells, and your solution would be much nicer and simpler if it were correct. Do you want to correct it, or should I proceed?

estebanlm:
it was essential? I had no idea. 
Please go ahead with the correction, I have really little time this days (and I'm traveling to Argentina friday).

nicolas-cellier-aka-nice:
The main problem with `malloc `is that we must not forget to `free` in every possible failure case, or we'll get a leaking heap... So yes, I think it was easier to `alloca`.
I think that the author started to patch one place, then realized he had to repeat same job again and again. I'm not a big fan of macros in C, but this is a case where it can save time/bugs.
I will try to implement a solution based on yours tonight and keep a single macro instead of two.

jecisc:
@nicolas-cellier-aka-nice Was your version implemented/integrated?

Since there is no reference to this PR it's hard to know if it was redone.

nicolas-cellier-aka-nice:
https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/162

jecisc:
Thank you for the info!

PR #166, opened 2017-11-03
Enable ASLR and DEP in Win32

ghost:
To increase the security of native code running in the VM against buffer overflows and other attacks, Address Space Layout Randomisation (ASLR) and Data Execution Prevention (DEP) can be enabled.

This pull request changes Makefile.tools for win32 to enable ASLR and DEP.

I've modified the PE header of Squeak.exe locally and it works.

### DIFF
--- a/build.win32x86/common/Makefile.tools
+++ b/build.win32x86/common/Makefile.tools
@@ -67,7 +67,7 @@ LD:=	 $(CC)
 LDCXX := $(TOOLPREFIX)clang++ # For linking c++ bundles
 
 EXPORT:=--export-all-symbols
-BASELDFLAGS:=-m32 -mthreads -Wl,--large-address-aware,$(EXPORT)
+BASELDFLAGS:=-m32 -mthreads -Wl,--large-address-aware,--dynamicbase,--nxcompat,$(EXPORT)
 LDFLAGS:= -mwindows $(BASELDFLAGS)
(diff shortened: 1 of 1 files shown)


nicolas-cellier-aka-nice:
Sounds reasonnable., I presume that the flags enable software DEP.
Would activation of DEP still be enabling the JIT?
(I understand that activation of DEP is configured somewhere else in the system, so it should not prevent inclusion of this PR)

ghost:
Yes, my VM has ASLR and DEP on and it works.

nicolas-cellier-aka-nice:
My question was probably completely irrelevant because DEP honours the executable page flags, and we explicitly ask for executable pages for the JIT don't we?

ghost:
Yes, you do, because otherwise my changed VM would crash.

nicolas-cellier-aka-nice:
Hi Luis, I would have expected to be redirected to usage of PAGE_EXECUTE_* macro or sqMakeMemoryExecutableFromTo() in initializeCodeZoneFromupTo() rather than just it works for me ;)

nicolas-cellier-aka-nice:
What about win64?

ghost:
Line 207 in sqWin32SpurAlloc.c contains PAGE_EXECUTE_READWRITE in sqMakeMemoryExecutableFromTo.

PR #167, opened 2017-11-07
Improve OpenBSD support (base files)

ckeen:
Hi,

These rather simple patches enable the compilation of the base source files. The build script adaption will happen in a separate pull request as this might require further refactoring of the existing build infrastructure so as not to break other platforms.

Note, that I have not tested the i386 related patches due to lack of HW. They do look Ok though.

### DIFF
--- a/platforms/unix/vm/sqUnixExternalPrims.c
+++ b/platforms/unix/vm/sqUnixExternalPrims.c
@@ -61,7 +61,11 @@
  
 #include <sys/param.h>
 #include <sys/stat.h>
-#include <sys/errno.h>
+#ifdef __OpenBSD__
+#   include <errno.h>
+#else
+#   include <sys/errno.h>
+#endif
 
 /* get a value for RTLD_NOW, with increasing levels of desperation... */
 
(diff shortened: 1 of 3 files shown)


ckeen:
I have no idea how to read those Travis logs but it looks like that is not related to my changes?

nicolas-cellier-aka-nice:
For the linux32 pharo vm, it's libgit2 compilation that fails with

>  error: size of array ‘__curl_rule_01__’ is negative

see https://travis-ci.org/OpenSmalltalk/opensmalltalk-vm/jobs/298742876
and https://stackoverflow.com/questions/35181744/size-of-array-curl-rule-01-is-negative

For the linux32 squeak VM, it seems that the test phase stale

> Running suite "Squeak-5.1 on TravisCI (873.9)" with 3699 tests.
...snip...
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

https://travis-ci.org/OpenSmalltalk/opensmalltalk-vm/jobs/298742871

What we need to check is the build status before the PR. I'm not sure it was ever green for a long time... (that ---biiiiiiiiiip--- !!!)

nicolas-cellier-aka-nice:
Hmm I just reviewed the changes, there's not a chance out of a million that it would change the linux build status, so I'm accepting immediately

PR #168, opened 2017-11-09
Feature lowcode linux build scripts update

ronsaldo:
Updating the Linux Lowcode VM building scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,10 @@ matrix:
         branch: Cog
         repo: OpenSmalltalk/opensmalltalk-vm
   - os: linux
+    env: ARCH="linux64x64" FLAVOR="pharo.cog.spur.lowcode" HEARTBEAT="threaded"
+  - os: linux
+    env: ARCH="linux64x64" FLAVOR="pharo.cog.spur.lowcode" HEARTBEAT="itimer"
+  - os: linux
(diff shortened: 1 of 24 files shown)


ronsaldo:
I see this error. Do not integrate yet.
libtool: link: gcc -m64 -shared  -fPIC -DPIC  .libs/SDL2DisplayPlugin.o   -lSDL2  -m64 -O2 -msse2 -Wl,-z -Wl,now   -Wl,-soname -Wl,SDL2DisplayPlugin.so -o .libs/SDL2DisplayPlugin.so

/usr/bin/ld: cannot find -lSDL2

ronsaldo:
I see that the SDL2DisplayPlugin is not being built on Linux since ages. I will try to fix this problem in a separate issue and pull request.

nicolas-cellier-aka-nice:
So should we wait a bit for integration?
(IMO there's low risk of collision with other devs anyway)

ronsaldo:
Not anymore :), if possible integrate it now. As I said, I will try to fix the plugin problem in a separate issue/PR. I will probably have to contact Esteban. It looks like I am not the only one interested on this.

krono:
@ronsaldo can you merge from `Cog` or Rebase so we see whether it still works? then we could merge.

krono:
@ronsaldo any news here?

eliotmiranda (approved):

PR #174, opened 2017-11-24
openssl: Fix compability with OpenSSL 1.1 as used by debian

zecke:
Norbert reported that a build for Debian9 fails to open outgoing
TLS connections. In OpenSSL the semantic of BIO_write for zero
sized writes has been changed. Do not attempt to write if the
length is zero. Check for the length before calling BIO_write.

I have updated the outgoing connection code and updated the
incoming connection one as well (didn't manually test it though)
and left the decrypt routine as is.

The Pharo testcase:
./products/phcogspurlinuxht/pharo ~/Pharo-60520.image eval "ZnEasy getPng: 'https://pharo.org/files/pharo.png'"

OpenSSL-issue: https://github.com/openssl/openssl/issues/4782


zecke:
@estebanlm

PR #179, opened 2017-12-26
Normal way to #include <errno.h>

aryler:
I'm pretty certain that
#include <errno.h>
is the right way to do it everywhere.

### DIFF
--- a/platforms/iOS/vm/OSX/sqMacUnixExternalPrims.m
+++ b/platforms/iOS/vm/OSX/sqMacUnixExternalPrims.m
@@ -54,7 +54,7 @@ extern SqueakOSXAppDelegate *gDelegateApp;
 #include <dlfcn.h> 
 #include <sys/param.h>
 #include <sys/stat.h>
-#include <sys/errno.h>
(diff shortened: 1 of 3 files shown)


nicolas-cellier-aka-nice:
It sounds correct, I verified `man errno` for freebsd, openbsd, linux and all mention `<errno.h>`
I think it comes from macos, where `man errno` does not answer anything usefull...
But where I get this answer to `man 2 intro`

**NAME**
     intro -- introduction to system calls and error numbers

**SYNOPSIS**
     #include <sys/errno.h>

**DESCRIPTION**
     This section provides an overview of the system calls, their error returns, and other common
     definitions and concepts.

aryler:
MacOS seems to be the only one that specifies sys/errno.h in documentation.
But if that were necessary then I suspect sys/errno.h would be found in more places in the codebase.
Even in the MacOS dir there are plenty of plain errno.h's.
Plain errno.h is POSIX.
I'm pretty certain it's normal for errno.h to #include sys/errno.h.

PR #185, opened 2018-01-03
fix SqueakSSL plugin build for PharoVM macOS

estebanlm:
at least in high sierra this was failing


krono:
What is openssl necessary for?

estebanlm:
No idea, but SqueakSSL needs it. 
then, we need it too (for other stuff like ssh2 and libgit2, etc.)

krono:
No, SqueakSSL on macOS does _not_ use openssl. In fact, the `-lssl -lcrypto` should be _removed_

krono:
If you need openssl for ssh2 and libgit2, Ok, but SqueakSSL is not the place for that

estebanlm:
ah well... it was like that (i didn't added the -lssl -lcrypto), and because of them it was failing the build. 
I just made sure compilation succeeds.

krono:
Yeah, I see, not meant to blame you.
I think I'll just remove it directly in `Cog`, ok?

eliotmiranda:
> On Jan 3, 2018, at 10:20 AM, Tobias Pape <notifications@github.com> wrote:
> 
> Yeah, I see, not meant to blame you.
> I think I'll just remove it directly in Cog, ok?
> 

Yes please :-)

krono:
It has been done :D

nicolas-cellier-aka-nice:
I presume that the changes for building openssl are still necessary (if openssl is used by some other libs).
Esteban, could you remove changes to platforms/iOS/plugins/SqueakSSL/Makefile and commit again?

estebanlm:
yeah, on his way

krono:
yay

PR #186, opened 2018-01-03
HowToBuild doesn't have to mention missing libssl

nicolas-cellier-aka-nice:
Because we don't need to link against libssl.
Keep the hint from Eliot for an hypothetical libfoobar.dylib missing.

### DIFF
--- a/build.macos32x86/HowToBuild
+++ b/build.macos32x86/HowToBuild
@@ -172,23 +172,28 @@ enough to build the system.  The 32-bit build requires at least the 10.9 SDK.
 To obtain SDKs it is best to web search for "obtaining older sdks for Xcode",
 and follow the instructions you find.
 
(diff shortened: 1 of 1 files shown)


fniephaus:
I don't think you need to open PRs for edits like this unless you want feedback of course.
Also, if you add `[ci skip]` to your commit message, you don't trigger a new CI build which is handy when you didn't edit code :)

Best,
Fabio

nicolas-cellier-aka-nice:
Thanks Fabio, [ci skip] is usefull
I edited from Web interface and I think it forced me to branch...
Let's ask Eliot's feedback explicitely (since he wrote the original)

fniephaus:
No worries! Thanks for taking care of this

krono:
Long sitting here. Let's do it.

PR #189, opened 2018-01-04
Let Win64 ABI use an unwind-free setjmp

nicolas-cellier-aka-nice:
More information can be found in this commit comment:
e45e7eb


nicolas-cellier-aka-nice:
More information can be found in this commit comment:
https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/e45e7ebf8f7baf4548990a6f7fa09c480cfbc683

eliotmiranda (approved):
This is good.  It matches what we already do in the interpreter.

PR #205, opened 2018-01-30
[SqueakSSL] Overlay OpenSSL for linux/unix

krono:
When we dynamically link against OpenSSL, the bundles are not portable,
as CentOS and friends use other SO_NAMEs than Debian and friends. Also,
soft-fallback for later features such as host name verification is hard.

When we statically link, we might lack behind the OS, the binaries are
bigger, and the legal situation is less clear.

So we now support not linking at all but rather lookup all necessary
functions/symbols at runtime.

This can be disabled with SQSSL_OPENSSL_LINKED which effectively results
in the dynamically-linked behavior. (This is preferable for platform
builds, eg, debs and rpms)

### DIFF
--- a/platforms/unix/plugins/SqueakSSL/openssl_overlay.h
+++ b/platforms/unix/plugins/SqueakSSL/openssl_overlay.h
@@ -169,7 +169,9 @@ void (*sqo_sk_pop_free)(STACK *st, void (*func) (void *));
 #define sqo_sk_GENERAL_NAME_pop_free(st, free_func)     \
   sqo_SKM_sk_pop_free(GENERAL_NAME, (st), (free_func))
 
-
+#if !defined(X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS)
+#define X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS 0x10
+#endif
 
 bool loadLibrary(void)
 {
--- a/platforms/unix/plugins/SqueakSSL/openssl_overlay.h
+++ b/platforms/unix/plugins/SqueakSSL/openssl_overlay.h
@@ -1,0 +1,299 @@
+#ifndef SQ_OPENSSL_OVERLAY_H
+#define SQ_OPENSSL_OVERLAY_H 1
+
+#include <stdbool.h>
+#include <dlfcn.h>
+#include <openssl/ssl.h>
+#include <openssl/err.h>
+#include <openssl/x509v3.h>
+
+
+static void* dlhandle_self = NULL;
+static void* dlhandle_crypto = NULL;
+static void* dlhandle_ssl = NULL;
+
+#ifndef __has_attribute
+#define __has_attribute(x) 0
+#endif
+
+#if __has_attribute(constructor) \
+  || defined(__GNUC__)
+#define SQO_CONSTRUCTOR __attribute__((constructor))
+#else
+#define SQO_CONSTRUCTOR /**/
+#endif
+
+void SQO_CONSTRUCTOR fini(void)
+{
+  if (dlhandle_self) { dlclose(dlhandle_self); }
+  if (dlhandle_crypto) { dlclose(dlhandle_crypto); }
+  if (dlhandle_ssl) { dlclose(dlhandle_ssl);}
+}
+
+#if (defined(DEBUG) || defined(DEBUGVM)) && !defined(NDEBUG)
+#define SQO_HAS_FOUND_SYM(s,n,h)                                \
+  do {                                                          \
+    s = dlsym(h, n);                                            \
+    if (s) {                                                    \
+      fprintf(stderr, "Found symbol %s in " #h "\n", n);        \
+      return s;                                                 \
+    }                                                           \
+  } while (0)
+#else
+#define SQO_HAS_FOUND_SYM(s,n,h)                \
+  do {                                          \
+    if ((s = dlsym(h, n))) {                    \
+      return s;                                 \
+    }                                           \
+  } while (0)
+#endif
+
+#if !defined(SQO_DL_FLAGS)
+#define SQO_DL_FLAGS RTLD_NOW | RTLD_GLOBAL | RTLD_NODELETE
+#endif
+
+#define SQO_FIND_SYM(sym, name, where, dlname)                  \
+  do {                                                          \
+    if (!dlhandle_ ## where) {                                  \
+      dlhandle_ ## where = dlopen(dlname, SQO_DL_FLAGS);        \
+      if (!dlhandle_ ## where) {                                \
+        fprintf(stderr, "cannot load from " #dlname "\n"); \
+        abort();                                                \
+      }                                                         \
+    }                                                           \
+    if (dlhandle_ ## where) {                                   \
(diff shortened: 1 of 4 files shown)


fniephaus:
Thanks, @krono! These are important SqueakSSL plugin updates and we might want them in the release which is due in two days. I'll give everyone 12 hours from now to review the code changes. If no one objects, I will merge this in once CI has passed.

@krono Could you please provide some concrete examples that can be used to test these changes from inside an image, so it's easier to verify these changes on all platforms?

/cc @eliotmiranda @estebanlm

platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.c
```diff
@@ -432,19 +438,20 @@ sqInt sqConnectSSL(sqInt handle, char* srcBuf, sqInt srcLen, char *dstBuf, sqInt
...
-			/* Try IP first, expect INVALID_IP_STRING to continue with hostname */
-			matched = (enum sqMatchResult) X509_check_ip_asc(cert, ssl->serverName, 0);
-			if (matched == INVALID_IP_STRING) {
-				matched = (enum sqMatchResult) X509_check_host(cert, ssl->serverName, serverNameLength, X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS, NULL);
-			}
-#else
-			matched = sqVerifyIP(ssl, cert, ssl->serverName, serverNameLength);
-			if (matched == INVALID_IP_STRING) {
-				matched = sqVerifyDNS(ssl, cert, ssl->serverName, serverNameLength);
+                        //#ifdef X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS
```
fniephaus: Is this comment/ifdef still needed?

platforms/unix/plugins/SqueakSSL/openssl_overlay.h
```diff
@@ -0,0 +1,301 @@
...
+#define sqo_sk_GENERAL_NAME_num(st)             \
+  sqo_SKM_sk_num(GENERAL_NAME, (st))
+#define sqo_sk_GENERAL_NAME_value(st, i)        \
+  sqo_SKM_sk_value(GENERAL_NAME, (st), (i))
+#define sqo_sk_GENERAL_NAME_free(st)            \
+  sqo_SKM_sk_free(GENERAL_NAME, (st))
+#define sqo_sk_GENERAL_NAME_pop_free(st, free_func)     \
+  sqo_SKM_sk_pop_free(GENERAL_NAME, (st), (free_func))
+
+#if !defined(X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS)
```
fniephaus: Could you add a comment when this is not defined? Otherwise, prefix it as well?
-> `sqo_X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS`

platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.c
```diff
@@ -275,14 +275,20 @@ sqInt sqCreateSSL(void) {
 	sqInt handle = 0;
 	sqSSL *ssl = NULL;
 
-	SSL_library_init();
-	SSL_load_error_strings();
+	if (!wasInitialized) {
+ 		if (!loadLibrary()) {
+			return 0;
+		}
+                sqo_SSL_library_init();
```
fniephaus: Fix indentation

platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.c
```diff
@@ -275,14 +275,20 @@ sqInt sqCreateSSL(void) {
...
+                sqo_SSL_load_error_strings();
```
fniephaus: Same

platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.c
```diff
@@ -95,25 +95,25 @@ enum sqMatchResult sqVerifyDNS(sqSSL* ssl, X509* cert, const char* serverName, c
...
-			const GENERAL_NAME* sAN = sk_GENERAL_NAME_value(sANs, i);
+			const GENERAL_NAME* sAN = sqo_sk_GENERAL_NAME_value(sANs, i);
 			if ((sAN->type == matchType) &&
 			    sqVerifySAN(ssl, sAN, serverName, serverNameLength, matchType)) {
 				matchFound = MATCH_FOUND;
 				break;
 			}
 		}
-		sk_GENERAL_NAME_pop_free(sANs, GENERAL_NAME_free);
+		sqo_sk_GENERAL_NAME_pop_free(sANs, (void(*)(void*))sqo_sk_free);
```
fniephaus, line 116: I don't understand what's going on here, but please double check if `(void(*)(void*))sqo_sk_free` is the correct substitution for `GENERAL_NAME_free` here.

platforms/unix/plugins/SqueakSSL/openssl_overlay.h
```diff
@@ -0,0 +1,301 @@
...
+  _C(sqo_sk_pop_free = (void (*)(_STACK *st, void (*func) (void *))) _sqo_find("sk_pop_free"));
+#else
+  _C(sqo_sk_new_null = (STACK *(*)(void)) _sqo_find("sk_new_null"));
+  _C(sqo_sk_push = (int (*)(STACK *st, char *data)) _sqo_find("sk_push"));
+  _C(sqo_sk_free = (void (*)(STACK *st)) _sqo_find("sk_free"));
+  _C(sqo_sk_value = (char *(*)(STACK *st, int i)) _sqo_find("sk_value"));
+  _C(sqo_sk_num = (int (*)(STACK *st)) _sqo_find("sk_num"));
+  _C(sqo_sk_pop_free = (void (*)(STACK *st, void (*func) (void *))) _sqo_find("sk_pop_free"));
+#endif // OPENSSL_VERSION_NUMBER >= 0x10000000L
+  return true;
```
fniephaus: Can't this be in the previous `#if OPENSSL_VERSION_NUMBER >= 0x10000000L` in line 282?

platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.c
krono: Because some expansion is taking place here that cannot be easily emulate. However, all `..._free` in the stack sense ultimately end up _being_ `sqo_sk_free`. Qt does it similarly.

platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.c
krono: No, its just a reminder.

fniephaus: It doesn't remind me of anything, maybe add a comment? ;)

krono:
Here's the most simple use case:

```Smalltalk
(WebClient httpGet: 'https://danluu.com') content
```

fniephaus:
With the help of @krono, I was able to verify that this is working fine for squeak.cog.spur on CentOS (x64 and i386).

PR #219, opened 2018-02-20
Added pharo.sista.spur builds

clementbera:
I checked the Mac 32 builds were working fine.
I did not check Linux & win.
Is there something to change on the CI to make them build ?


clementbera:
I am not sure what is the process from there so I'll wait for some one like Fabio to comment the pull request.

clementbera:
I am a bit confused here. Builds is red but I did not change anything ?

fniephaus:
Hi Clément,
I'm not sure why the builds were failing, but I've just pushed a commit that add `pharo.sista.spur` builds to our CI. Let's see if the builds are passing this time...

clementbera:
64 bits sista builds can be disabled for now. So it looks good.

I wonder, where do I need to change things so that the build upload the built VM into file.pharo.org/vm/sista in a similar way to the other VMs ? That's the only missing thing.

fniephaus:
Ok cool, I've just disabled the 64bit builds.
Please get in touch with @estebanlm with regard to the file uploads. Right now, I assume the sista builds are uploaded to the file server, but may override existing files depending on the build order. The reason for that might be that the pack/deploy scripts for Pharo do not know about sista yet and treat all VMs as a standard spur vm.

clementbera:
Let's merge I will patch deployment later.

fniephaus:
@estebanlm and I have fixed the problem described above in the meantime. Hopefully, everything works and you don't have to patch anything :)

PR #224, opened 2018-03-07
Do not cache fileSize in SQFile

smalltalking:
If you have more than one reference to the same file from the image (e.g. the changes file), then the file size cache will prevent changes written by one reference to be noticed by another, so you have to reopen all other references to actualize their cached file size.  
These changes will make that unnecessary.   

I could only test the code on linux, but hopefully the MacOS/iOS and win32 implementations are correct too.  
The changes do not cover platforms Plan9 and RiscOS. That's up to someone who has access to and experience with them.  
There's room for optimization on platforms which support the fstat() function (the current implementation just seeks to the end of the file and then back to the original position).


akgrant43:
Hi @smalltalking,

I'm just wondering why you haven't followed up on this PR?

The linux builds failed because there is still a reference to setSize in
sqFilePluginBasicPrims.c.

When I tested the speed difference, not caching the size added about 2
microseconds to retrieving the file size on my 4 or 5 year old Dell
XPS13 (SSD).

Thanks,
Alistair

smalltalking:
Hi @akgrant43,

I wrote these changes last October. There was a small discussion about them [on the mailing list](http://forum.world.st/About-SQFile-s-fileSize-field-td4981131.html).
I wanted to create a pull request in October, but I had some other things to do.
After I made the pull request, I saw that the changes interfered with your FilePlugin changes, so I decided to wait till those are fully integrated.

akgrant43:
Thanks for the explanation and link to the email discussion.

The remaining changes I have planned for FilePlugin will take quite a while (I need to learn about the simulator first).

Can you update the PR and I'll merge it in?

Thanks,
Alistair

PR #226, opened 2018-03-11
The Yakshaving

krono:
Ok, don't mind the branch name. This started out as a trial to improve OpenBSD support, especially since my SqueakSSL/Unix changes broke it for them. But things got out of hand (viz. [Yak](https://en.wiktionary.org/wiki/yak_shaving) [shaving](https://www.urbandictionary.com/define.php?term=yak%20shaving) )

## Travis Changes

 - This adds a few more libraries for Travis, such that more plugins are actually built and not disabled due to missing dependencies

## `scripts/updateSCCSVersions`

 - This is a `bash` script now, but `bash` is not in `/usr/bin` on all platforms. Solution: use `/usr/bin/env bash`. Problem 2: We want `bash -e`, but POSIX/execve only specifies `#! <interpreter> <onearg>`. Solution 2:

      ```shell
      #! /usr/bin/env bash
      set -e
      ```
    @fniephaus we should do that for all instances of `bash -e`, probably.

## Autoconf.

- A lot of selections, flags, and specifics where handled in dispersed `mvm` files.
- Most of which can now be handled via the _results_ of autoconf, that is
  - the `Makefile`s (with their respective flags)
  - the `config.h`
- Also, modernized a good bunch of autoconf/m4 files to more recent styles.
- And, finally, selection of OpenSSL/LibreSSL backend settings for SqueakSSL via `configure`.

### Missing here
- not all `acinclude.m4` adhere to autoconf's current style.
- Cog's options (eg, `COGMTVM`) are not yet covered.

## Tiny things

 - Improve no-op definotion of `VM_LABEL` as not to make `gcc` complain about strange expressions.


platforms/unix/plugins/SqueakSSL/sqUnixSSL.c
```diff
@@ -0,0 +1,18 @@
+/* -*- mode: c; -*- */
+/****************************************************************************
+ *   PROJECT: SqueakSSL implementation for Unix
+ *   FILE:    sqUnixSSL.c
+ *   CONTENT: SSL platform functions
+ *
+ *   AUTHORS: Tobias Pape (topa)
+ *               Hasso Plattner Institute, Postdam, Germany
```
j4yk: typo ;-)

krono: GOD DAMMIT

PR #213, opened 2018-02-05
Make updateSCCSVersions work in more situations

j4yk:
- allow spaces in paths (added quotes)
- support cases where the repository is not in a .git directory at the top of the working tree
- do not hardcode the directory level of this script
- avoid stashing the whole working tree when only two well-known files need to be updated
- touch should suffice to make a file different from its index version (i. e., eligible for checkout)

Please review for compatibility and pick those parts of the changes that you like.

PR #227, opened 2018-03-11
Second attempt to improve updateSCCSVersions

j4yk:
I finally got around to solve the "not a git repository" issue with my previous pull request. For the analysis see below. Please review and try it out before merging.

When the script is run as a hook, the environment variable `GIT_DIR` is set to `.git` (path relative to the working tree) for regular repositories. This affects the behavior of `rev-parse`. When the script is run from the working tree and the checkout commands towards the bottom are executed, the script actually invokes itself as the post-checkout script. During that inner invocation, the script would `cd` to `.git/hooks`, and because of the GIT_DIR variable, `rev-parse` would then look for the git repository in `.git/hooks/.git`. This failed, of course.
The second error message "unrecognized input" came from `git apply` when it was fed with empty input (because there were no uncommitted changes to either *Version.h file).

Since GIT_DIR caused the trouble, why not use it to detect whether the script is run as a hook...


krono:
@j4yk, can you tell me whether this is still relevant?

j4yk:
Well I guess so for the benefits outlined in #213. But I cannot check now whether anything has changed in the baseline script in the meantime.

Has anyone tried it, to avoid that it only works on my machine?

krono:
@j4yk can you rebase onto Cog or merge Cog into here? let's see to get this going

nicolas-cellier-aka-nice:
Hi Tobias, I tried the resolve thru github which is a bit scary but OK for such simple change.
The sole conflict was your own change of a tab to spaces...

fniephaus:
Now that this has been merged, do not forget to run the script manually once in order to update your local hooks. Otherwise, your local copy will still be using `git stash`.

krono:
@fniephaus We should ping vm-dev regarding that

eliotmiranda:
First, Jakob, huge thanks for this.  The weird symptom of getting merge conflicts appearing in certain files after each push has disappeared.  Second, I did apply the script manually but it made no alterations to sqSCCSVersion.h in my case (it was already localized to my machine).  So it may not be necessary to run it, but it does no harm.

fniephaus:
> Second, I did apply the script manually but it made no alterations to sqSCCSVersion.h in my case (it was already localized to my machine). So it may not be necessary to run it, but it does no harm.

The behavior with regard to `sqSCCSVersion.h` shouldn't change at all. More importantly, running the script manually also copies and, in our case, updates the script in the repository's `hooks` directory (see [here](https://github.com/j4yk/opensmalltalk-vm/blob/623dee2b6589723173af8b10d73485a23202c44f/scripts/updateSCCSVersions#L23)) to configure Git for our needs. This change is within the `.git` directory and therefore doesn't alter your local copy, the OpenSmalltalkVM code base.

PR #228, opened 2018-03-15
Fixed getGoodSpur64VM.sh: 64-bit VM directory name is different

smalltalking:
mv quit with an error, because the directory didn't exist.


smalltalking:
Should have tagged it to avoid invoking CI...

PR #230, opened 2018-03-22
Fixed typo in the download URL

smalltalking:


clementbera:
Seems to solve a typo without side-effect - Travis is not showing any new problems even if red.

krono:
Yes, this script is not used by travis.

PR #231, opened 2018-03-25
macos: Fix --trace for the PharoVM

zecke:
Introduce VMOPTIONLEN that can offset the extra '-' in the
calculations.

### DIFF
--- a/platforms/iOS/vm/OSX/sqSqueakOSXApplication.m
+++ b/platforms/iOS/vm/OSX/sqSqueakOSXApplication.m
@@ -65,9 +65,11 @@
 #ifdef PharoVM
 # define VMOPTION(arg) "--"arg
 # define VMOPTIONOBJ(arg) @"--"arg
+# define VMOPTIONLEN(n) (1+n)
 #else
 # define VMOPTION(arg) "-"arg
 # define VMOPTIONOBJ(arg) @"-"arg
+# define VMOPTIONLEN(n) (n)
 #endif
 
 usqInt gMaxHeapSize = 512*1024*1024;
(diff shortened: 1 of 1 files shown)


platforms/iOS/vm/OSX/sqSqueakOSXApplication.m
```diff
@@ -166,7 +168,7 @@ - (int) parseArgument: (NSString *) argData peek: (char *) peek
 	/* Options with no arguments */
 
 	NS_DURING;
-		if ([argData compare:  VMOPTIONOBJ("psn_") options: NSLiteralSearch range: NSMakeRange(0,5)] == NSOrderedSame) {
+		if ([argData compare:  VMOPTIONOBJ("psn_") options: NSLiteralSearch range: NSMakeRange(0,VMOPTIONLEN(5))] == NSOrderedSame) {
```
krono: Side comment: using `VMOPTIONOBJ` makes no sense for "psn_". It will always have exactly one dash, so using VMOPTIONOBJ (and the otherwise useful VMOPTIONLEN) makes no sense

zecke: Good point. OSX doesn't care about the number of dashes we have. I don't like the code. We should strip the dashes before parsing the arguments but that would have been a bigger change and outside the timebox I had. :(

krono: No problem, just a side comment to not forget

nicolas-cellier-aka-nice:
Agree, we can do better, but this is good enough, we'll see better later.
See also
https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/136
https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/181

PR #233, opened 2018-04-02
vm version string: add commit short hash

akgrant43:
Modify the version string so that the git commit short hash is displayed
after the commit date.


fniephaus:
+1 for Git commit short hashes in the version string.

PR #238, opened 2018-04-06
Do not convert line ending of image/*.sh files on windows

jecisc:
Cygwin cannot run scripts with crlf. I added image/*.sh to the gitattributes to not convert their line ending.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,7 @@
 # Bash needs lf even on windows/cygwin
 /scripts/* eol=lf
 /image/*.st eol=lf
+/image/*.sh eol=lf
 mvm eol=lf
 sq*SCCSVersion.h filter=RevDateURL ident
 *.changes -diff


akgrant43:
Hi Cyril,

Which file was causing problems (so I can do a git checkout and manual inspect the file before applying your PR)?

Thanks,
Alistair

jecisc:
It was buildspurvmmakerimage.sh and all the scripts it calls.

nicolas-cellier-aka-nice:
Note that there is a workaround: put `set -o igncr` in your .bash_profile
See https://stackoverflow.com/questions/14598753/running-bash-script-in-cygwin-on-windows-7 for example.

I agree that it's preferable to correctly set the git file attributes and avoid the problem at the source!

jecisc:
Thank you for the information! I did not knew. I'll probably still try to avoid adding this, like that I'll encounter the problems and do PR to correct them.

PR #239, opened 2018-04-06
Add instructions to download squeak vm for cygwin to build VMMaker

jecisc:
In the image directory, there are scripts to prepare a VMMaker image with squeak.

Those scripts have instructions to download the latest squeak vm depending on the OS. Currently, there is only Darwin and Linux. This PR add instructions for Cygwin.


clementbera:
Merged, harmless, script related.

PR #253, opened 2018-04-18
another try to properly build libgit2

estebanlm:


estebanlm:
will create another PR

PR #254, opened 2018-04-24
Add accurate check for console stdio + add associated primitive 

VincentBlondeau:
Related to https://github.com/pharo-project/pharo/pull/694


eliotmiranda:
Hi Vincent, in isStdioDescriptorATTY at line 898 there is no guard to check if pGetFileInformationByHandleEx has already been fetched.  The code needs to be something like

	static pfnGetFileInformationByHandleEx pGetFileInformationByHandleEx = NULL;

	if (! pGetFileInformationByHandleEx) {
		pGetFileInformationByHandleEx = (pfnGetFileInformationByHandleEx)
 			GetProcAddress(GetModuleHandle(TEXT("kernel32.dll")),
						"GetFileInformationByHandleEx");
		if (! pGetFileInformationByHandleEx)
			return -2;
	}

Also, a word to the wise.  It's great to save vertical real estate putting things on long lines, but not for error returns, because of line oriented debugging.  We may want to put a breakpoint on an error return, not the guard for that return.  And so, the error return should be on a different line to the guard.

eliotmiranda:
Also, in platforms/Cross/plugins/FilePlugin/sqFilePluginBasicPrims.c, the comment "//There is always an TTY to write into for Unix and Mac" (and hence the implementation) is wrong.  It should be something like

```
sqInt sqStdioDescriptorIsATTY(void) {
	return isatty(fileno(stdin));
 }
```

VincentBlondeau:
Thanks Eliot for the piece of advice! I am not an expert in C yet ;)

I made the changes but I was wondering why it is needed to set a guard: the line before just set the pGetFileInformationByHandleEx to NULL.

akgrant43:
Hi Vincent,

Just based on a quick visual inspection of the code:

- It looks like on Unix sqStdioDescriptorIsATTY() checks stdin, while on Windows it checks stdout, which is a bit inconsistent and they can be redirected independently anyway.  Why not pass an argument allowing the user to specify which file to check (0, 1, 2)?
- We'll want to modify sqFileStdioHandlesInto() in sqWin32FilePrims.c to use the new check (instead of GetConsoleMode()).
- Actually, since the result is always stored in the SQFile structure for the stdio streams, the primitive could just return the cached value rather than determining it every time.
- sqFileAtEnd() on Windows just assumes that stdin is never at EOF if it is a terminal.  I don't know how EOF is signalled in a terminal on Windows (Ctrl-D?), but we could also improve this (although maybe this can be a separate PR).

Thanks!
Alistair

eliotmiranda:
Hi Vincent,  shouldn't all the
    files[0].isStdioStream = sqIsFileDescriptorATTY(0);
occurrences in platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c: sqFileStdioHandlesInto read either
    files[0].isStdioStream = sqIsFileDescriptorATTY(STD_INPUT_HANDLE);
or
    files[0].isStdioStream = sqIsFileDescriptorATTY(files[0].file);
?  Since sqIsFileDescriptorATTY could a HANDLE right?  And then you could get rid of the 
    h = (HANDLE)_get_osfhandle(fdNum);
call in sqIsFileDescriptorATTY, and rename it to sqIsFileHandleATTY.  yes it ,means that the Windows and Unix code differ a little, but they are different.

VincentBlondeau:
The exposed primitive sqIsFileDescriptorATTY should not receive an HANDLE because one needs to call sqIsFileDescriptorATTY to know if an handle is available. But I can change sqWin32FilePrims.c:isFileDescriptorATTY to take an HANDLE as argument instead of a String

akgrant43:
Hi Vincent,

After checking out and building VincentBlondeau:addStdoutIsConsolePrimitive I get:

```
~/pharo7/pharo-snap/pharo-vm/opensmalltalk-vm/build.win32x86/pharo.cog.spur/builddbg/vm
$ gdb PharoConsole.exe
GNU gdb (GDB) (Cygwin 7.10.1-1) 7.10.1
Copyright (C) 2015 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "i686-pc-cygwin".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from PharoConsole.exe...done.
(gdb) run --version
Starting program: /home/alistair/pharo7/pharo-snap/pharo-vm/opensmalltalk-vm/build.win32x86/pharo.cog.spur/builddbg/vm/PharoConsole.exe --                                               version
[New Thread 6252.0x16b0]
[New Thread 6252.0x1114]
[New Thread 6252.0x1a68]
[New Thread 6252.0x1d10]

Program received signal SIGSEGV, Segmentation fault.
0x76e1299b in msvcrt!_fileno () from /cygdrive/c/WINDOWS/System32/msvcrt.dll
(gdb) bt
#0  0x76e1299b in msvcrt!_fileno () from /cygdrive/c/WINDOWS/System32/msvcrt.dll
#1  0x00b4b17b in ?? ()
#2  0x00b4b333 in ?? ()
#3  0x00b4c567 in ?? ()
#4  0x00c22a7d in ?? ()
#5  0x00a913e3 in ?? ()
#6  0x74fc9564 in KERNEL32!BaseThreadInitThunk () from /cygdrive/c/WINDOWS/System32/KERNEL32.DLL
#7  0x773529ac in ntdll!EtwProcessPrivateLoggerRequest () from /cygdrive/c/WINDOWS/SYSTEM32/ntdll.dll
#8  0x77352980 in ntdll!EtwProcessPrivateLoggerRequest () from /cygdrive/c/WINDOWS/SYSTEM32/ntdll.dll
#9  0x00000000 in ?? ()
(gdb)
```

Sorry, but I'm out of time for today.  Hopefully I'll get to investigate a bit more over the coming days.

Cheers,
Alistair

VincentBlondeau:
I was using the wrong function to convert the handle to a number. I found another solution that works, now, I hope it will be working

akgrant43:
Hi Vincent & Eliot,

This still isn't working because 
sqFileStdioHandlesIntoFile_WithHandle_IsWritable() doesn't take SQFile 
by reference (arrays are just pointers to the first element in C, which 
is why the array of SQFile structures doesn't need to be passed by 
reference explicitly):


```
sqFileStdioHandlesIntoFile_WithHandle_IsWritable(SQFile *file, HANDLE handle, int isWritable) {
        file->sessionID = thisSession;
        file->file = handle;
        file->writable = isWritable;
        file->lastOp = 0; /* unused on win32 */
        file->isStdioStream = isFileHandleATTY(handle);
        AddHandleToTable(win32Files, handle);
}


sqFileStdioHandlesInto(SQFile files[3])
{
        sqFileStdioHandlesIntoFile_WithHandle_IsWritable(&files[0], GetStdHandle(STD_INPUT_HANDLE), false);
        sqFileStdioHandlesIntoFile_WithHandle_IsWritable(&files[1], GetStdHandle(STD_OUTPUT_HANDLE), true);
        sqFileStdioHandlesIntoFile_WithHandle_IsWritable(&files[2], GetStdHandle(STD_ERROR_HANDLE), true);

        return 7;
}
```


I think that gets the basic changes you wanted done.  But it still 
leaves stdio in cygwin terminals broken.  After a bit more reading it 
looks like cygwin terminals and Windows ReadConsole() and WriteConsole() 
functions are fundamentally incompatible.

I don't understand what ReadConsole() and WriteConsole() provide in 
terms of benefits over FILE* streams, so I can't really comment on 
the cost of replacing them.

Eliot, what are your thoughts about moving the stdio functionlity to use 
FILE* streams instead of HANDLE streams?

I think that would allow the VM to use fread(), feof(), and fwrite() to 
read and write stdio to terminals (cygwin), consoles (windows), pipes 
and regular files.

Another option may be to leave the existing console functionality and 
use the connectToFileDescriptor() routine to separately open the stdio 
streams as FILE* streams.  The image can then decide how it wants to 
interact with stdio.

(we should probably move this conversation to vm-dev and out of this 
particular PR)


Thanks,
Alistair

VincentBlondeau:
I forgot that structs are directly stored in the array without pointer to them. I fixed it.
What do you mean by broken? Do you have an example I can reproduce?

akgrant43:
Hi Vincent & Eliot,

> What do you mean by broken? Do you have an example I can reproduce?

Sure - if you start PharoConsole.exe from within a cygwin (mintty) 
terminal and remove the #isWin32 block from Stdio 
class>>standardIOStreamNamed:forWrite: so that you are actually using 
stdout, and then attempt to write something to stdout you'll get a 
primitive failure.  The same sequence works from a dos box (cmd.exe).

That's because the read and write primitives rely on the Windows 
ReadConsole() and WriteConsole() functions, which only work in a dos 
box, not in mintty.

When we started these modifications I assumed that it would be possible 
to have I/O that worked in both cygwin (mintty) and a dos box (cmd.exe).  
However on further reading it looks like MingW programs (which use 
either Windows functions such as ReadConsole() and WriteConsole() or 
the MS posix implementation) will only work properly in a dos box.  If 
you want proper mintty support you need to use the cygwin implementation 
of the posix functions.

As an example, attempting to use the MingW/MS fread() function and 
signal EOF:

- In a dos box, typing Control-Z correctly signals EOF.
- In a cygwin terminal (mintty), Control-D is never passed to the 
  program, so EOF is never true. (If you then type Control-C to 
  terminate the program, sometimes the terminal then consumes the 
  earlier Control-D and logs you out :-().

Someone please explain how I am wrong and how to get stdio working on 
both dos boxes and cygwin terminals...

Thanks,
Alistair

akgrant43:
Hi Vincent & Eliot,

> Ugh.  I didn't realize one couldn't use the same code in both :-(.  I guess we have to test to find out what the context is and use dread in one and ReadConsole in the other. 

I think it is worse than that.  If I use the following test code in a 
cygwin terminal:

```
#include <stdio.h>
#include <stdlib.h>


int main()
{
char	buf[1024];
char	*bufp;
int	cread;
int	count = 0;

bufp = &buf[0];
do {
    cread = fread(bufp, 1, 1, stdin);
    count += cread;
    bufp += cread;
    if (count > 16) break; }
while (cread > 0);
buf[count] = 0;

if (! feof(stdin))
    fprintf(stderr, "Error, not at end of file\n");

printf("--\n");
printf("%s\n", buf);
printf("--\n");
printf("Read %d characters\n", count);
exit(0);
}
```

And compile it with mingw:

```
$ i686-w64-mingw32-gcc -m32 consolestdio.c
```

It won't recognise EOF (Control-D).

```
$ ./a.exe
hello world
<Ctrl-D>
asdfasdfasdfasdfasdf
--
hello world

asdf
--
Read 17 characters
Error, not at end of file
```



If I compile it with gcc:

```
$ gcc -m32 consolestdio.c
```

It works as expected:

```
$ ./a.exe
hello world
<Ctrl-D>
--
hello world

--
Read 12 characters
```


So the best option I see for supporting both cygwin terminals and dos
boxes is to make our own dll that wraps around the cygwin1.dll functions
and to load it as required.  Although I'm a bit concerned that I haven't
found any mention of doing this elsewhere (and it's a lot of work for 
one specific use case).



> But I don't get it.  IIRC, it used to be the case that ReadConsole/WriteConsole worked in both.

I'd still like to be proved wrong...



Thanks,
Alistair

akgrant43:
Hi Ben,

> Try Ctrl-Z
> "In a non-Cygwin Windows program, Ctrl-Z on input triggers an end-of-file condition."

I had tried this, and since it's bash it just stopped the program, as expected.

I'd also looked at the page you referenced, but on reading it again thought I'd try unsetting the susp key:

```
stty susp ''
```

And sure enough, running the mingw version works this time:

```
$ ./mingw.exe
asdf<Enter>
<Ctrl-Z><Enter>
--
asdf

--
Read 5 characters
```

While it is a workaround, I'm not sure that asking users to modify the terminal settings and remember to use Ctrl-Z instead of Ctrl-D for Pharo in bash is going to be practical.

What do you think?

Just for LOLs, the following does work:

```
$ cat | vm/PharoConsole Pharo7.0.image eval StdioStreamTest manualStdinTest | tee /dev/null
asdf<Enter>
<Ctrl-D>
```

It doesn't work in headless mode, I haven't figured out why yet.


Cheers,
Alistair

eliotmiranda:
So this is real goodness.  Can we not address the issues and get this in?  Right now as I understand it we know how to discover both a Cygwin console and a Windows Console, but the two require separate approaches.  Can we not add both of them?  As far as the end-of-file issue dgoes I would say that is not a deal breaker.  If I were implementing a REPL I would want EOF to cause the system to exit cleanly but I would also live with having to issue a proper quit command.

akgrant43:
Hi Eliot and Vincent,

(I only have phone access, so apologies for typos, formatting, etc.)

Unless I'm misunderstanding the situation, we can't easily add support 
for both cygwin terminals and windows consoles.

We're compiling and linking with mingw, which means we have access to 
the windows consoles directly (through ReadConsole() and 
WriteConsole()).  If we want to access the cygwin terminals properly we 
need to use the posix io functions in cygwin1.dll (not the MS supplied 
versions of the posix io functions).

I guess it would be possible to write a wrapper DLL around the posix io 
functions in cygwin1.dll and load that if we recognise that we are in a 
cygwin terminal.  But that is obviously a fair bit of effort (not 
something that I'm interested in doing in the short or medium term).

Given your acceptance of not recognising EOF correctly in a cygwin 
terminal I would extend my previous suggestion that we modify Vincent's 
primitiveIsFileDescriptorATTY to return an enumerated value:

0 - no console (windows only)
1 - normal terminal (unix terminal / windows console)
2 - pipe
3 - file
4 - cygwin terminal (windows only)

What do you think?

Cheers,
Alistair

eliotmiranda:
Hi Alistair,

    I agree we shouldn't include cygwin1.dll.  But the existence of cygwin1.dll is proof that the relevant code can be implemented above Win32.  I do know that one can read from cygwin terminals; I was doing that regularly with the old VM.  So the issue is simply identifying a cygwin console as a tty.  We can put reading/writing from/to it to one side and first focus on identification.  I'm happy for you to extend primitiveIsFileDescriptorATTY as you suggest.  What I think we need, though, is to be able to identify either a Windows console or a cygwin console independently of cygwin1.dll (even if we have to copy some liberally licensed code from it).

akgrant43:
Great, I think we're in agreement.

Vincent, what do you think?

Cheers,
Alistair

VincentBlondeau:
Sounds good! The only thing is I can check the result only under windows. So, I don't know which values to return for Linux if the terminal is not a tty. Should I return 2 or 3? And how do I know that is a file or a pipe?

eliotmiranda:
Hi Vincent,

    Alistair defined the values earlier in the thread:

0 - no console (windows only)
1 - normal terminal (unix terminal / windows console)
2 - pipe
3 - file
4 - cygwin terminal (windows only)

I expect it'll be fine for you to extend the windows code but leave the Unix code for Alistair to implement.

akgrant43:
Hi Vincent,

I assumed that for Windows you could already distinguish between pipes and normal files since the pty check uses the pipe name as part of its check.  If not...

The fstat() st_mode field has S_IFIFO which indicates a first-in first-out (pipe) file.

Some of the documentation I saw said that S_IFIFO is for named pipes, I don't know what it will return for stdin being a pipe.

fstat() takes a file descriptor, so you'll need to convert a HANDLE to a fd.

If you can get Windows working I'm happy to take care of the Unix side.

Cheers,
Alistair

VincentBlondeau:
That should do it with these adds.
However, I do not know how to test if pipes or files are used under windows.

akgrant43:
Hi Vincent,

In cygwin you can test pipes with:

```
$ cat | PharoConsole Pharo7.0.image | tee /dev/null
```

It's also a hacky way to get terminal i/o working in a cygwin terminal.

In a windows console:

```
echo "hello world" | PharoConsole Pharo7.0.image
```

Should allow you to test stdin at least (untested).

In both cygwin terminals and windows consoles file i/o can be tested with:

```
PharoConsole Pharo7.0.image < a.txt > b.txt
```

HTH,
Alistair

VincentBlondeau:
That's helping!
Tested and working!

Thanks Alistair!

platforms/RiscOS/plugins/FilePlugin/sqFilePluginBasicPrims.c
```diff
@@ -407,6 +407,11 @@ sqInt sqFileStdioHandlesInto(SQFile files[3]) {
 	return 0;
 }
 
+sqInt sqFileDescriptorType(int fdNum) {
+	//Not implemented
+	return 0;
```
akgrant43: RiscOS should return -1 (error), not 0 (no console).

akgrant43 (commented):
One minor fix for RiscOS and this is good to go (I'll update the code for Unix after this has been integrated).

akgrant43 (approved):
Build failures are unrelated.

I'll wait 24 hours before merging in case anyone else wants to comment.

VincentBlondeau:
@akgrant43 : Are the changes I made ok to be integrated?

akgrant43:
Hi Vincent,

> Are the changes I made ok to be integrated?

Yes :-)  Sorry, I'm not getting much Pharo time at the moment.

Cheers,
Alistair

PR #255, opened 2018-05-07
another try to properly build libgit2 (v3)

estebanlm:


estebanlm:
this PR should be ready to integrate, but it is failing on sista builds (no related with my changes).
Con someone take a look?
Should I integrate anyway?

.travis_install.sh
```diff
@@ -43,7 +43,8 @@ elif [[ "${ARCH}" = "linux32x86" ]]; then
             libaudio-dev:i386 \
             gcc-multilib \
             uuid-dev:i386 \
-            libcurl3-dev:i386
+            libcurl4-openssl-dev:i386 \
+			libssh2-1-dev:i386
```
akgrant43, line 47: Are installing:

- libssh2-1-dev:i386
- libcurl4-openssl-dev:i386

an artefact of trying to figure out this issue, or really required?  If they are required, should the 64 bit libraries be added as well (and why isn't the 64 bit version of libcurl3 loaded? :-))

estebanlm: uhm, yes... those are remainders of me trying to solve the problem. 
probably not needed :)

build.linux32x86/third-party/Makefile.libgit2
```diff
@@ -56,6 +56,7 @@ install: $(THIRDPARTYLIBDIR)/$(LIBGIT2LIBNAME)
 			cp -d $$each $(THIRDPARTYINSTALLDIR); \
 		fi \
 	done
+	ldd $(THIRDPARTYLIBDIR)/$(LIBGIT2LIBNAME)
```
estebanlm, line 59: technically, this is not needed anymore either (I put it to see the bindings of libgit2 to ensure it was linking libssh2). Anyway since this can break again I would keep it.

akgrant43: Exactly why I didn't comment here. :-)

estebanlm:
meh, I'm going to accept it anyway, since sista builds are also failing in Cog branch

PR #261, opened 2018-05-15
.travis_install.sh update

akgrant43:
Run apt-get update to ensure we are using the current packages.

Hopefully this will remove the various

E: Failed to fetch http://something.deb  404  Not Found [IP: 35.184.213.5 80]

messages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: c
 
 sudo: required
 
+addons:
+    apt:
+      update: true
+
 cache:
   directories:
   - armchroot
--- a/.travis_install.sh
+++ b/.travis_install.sh
@@ -1,7 +1,7 @@
 set -e
 
-sudo apt-get update -y
 if [[ "${ARCH}" = "linux64x64" ]]; then
+    sudo apt-get update -y
     sudo apt-get install -yq --no-install-suggests --no-install-recommends --force-yes \
             debhelper \
             devscripts \
@@ -17,6 +17,7 @@ if [[ "${ARCH}" = "linux64x64" ]]; then
             gcc-multilib \
             uuid-dev
 elif [[ "${ARCH}" = "linux32x86" ]]; then
+    sudo apt-get update -y
     sudo apt-get remove -q -y gvfs-daemons
     sudo apt-get install -yq --no-install-suggests --no-install-recommends --force-yes \
             devscripts \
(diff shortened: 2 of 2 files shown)


Rydier:
Error logs of linux builds for
https://travis-ci.org/OpenSmalltalk/opensmalltalk-vm/jobs/378430866 
(and onwards) state they try loading 
http://security.ubuntu.com/ubuntu/pool/main/g/gcc-4.8/libx32itm1_4.8.4-2ubuntu1~14.04.3_amd64.deb
while the file on server is
http://security.ubuntu.com/ubuntu/pool/main/g/gcc-4.8/libx32itm1_4.8.4-2ubuntu1~14.04.4_amd64.deb
so an out-of-date package cache is the likely culprit?

krono:
Hmm, this seems like a travis problem then, too

akgrant43:
> so an out-of-date package cache is the likely culprit?

That's my guess, and why I tried this PR (which is admittedly experimental).

It would be good if someone could confirm that the changes make sense (I'm not familiar with travis at all).

Thanks,
Alistair

krono:
@fniephaus do you have an idea?

fniephaus:
I haven't seen those errors before, so I am also assuming that there was/is a problem on Travis' side. A quick look at they status page revealed [this incident](https://www.traviscistatus.com/incidents/fvch7h32ncf1) which mentions `apt-get update`-related problems.

@akgrant43 was this a temporary problem or does the problem still exist? And for how long have you seen these errors (hours/days/weeks)? 

Best,
Fabio

akgrant43:
Hi Fabio,

The apt-get update issues mentioned on the [Travis page](https://www.traviscistatus.com/incidents/fvch7h32ncf1) match the symptoms we're seeing.  My interpretation of what is written there is that this PR should be accepted now and reverted once Travis have their infrastructure issues sorted out.

The problem still existed as of a few hours ago.  It looks like the first instance was a couple of days ago, but the last commit before that was 6 days ago - about the time the Travis reports the issue as arising.

Cheers,
Alistair

fniephaus:
@akgrant43 why merge and roll back later? I mean, why is this just a temporary fix? What are the disadvantages/advantages if we keep the commit?

akgrant43:
> why merge and roll back later? I mean, why is this just a temporary fix? What are the disadvantages/advantages if we keep the commit?

The advantage is that the script will survive Travis not updating the environment for us.

The disadvantages are:

- It's a bit slower (there's obviously more being done).
- It increases slightly the chances that a temporary network problem will causes the build to fail.

Of course, there may be others that I haven't thought of.

I should also add that it isn't 100% clear to me that Travis will ever "fix" the issue.  They may decide that if you want to install software you should do an update - they mention trying to automatically determine whether an install is done, and maybe our script isn't detected.

Cheers,
Alistair

akgrant43:
There doesn't seem to be any real opposition to this, it's a small update that can easily be reversed, so if there's no dissenting comment in the next 30 minutes I'll merge it in.

Cheers,
Alistair

PR #266, opened 2018-06-06
BuildSqueakSpurTrunkVMMakerImage.st sort versions, use FileAttributes…

akgrant43:
…Plugin.oscog

BuildSqueakSpurTrunkVMMakerImage.st currently assumes that packages
returned from a repository are sorted by version, however this doesn't
appear to be true for smalltalkhub.com.

- Sort packages by version to ensure the latest is used.
- Load FileAttributesPlugin.oscog (which replaced FileAttributesPlugin)


akgrant43:
BuildSqueakSpurTrunkVMMakerImage.st isn't part of the build process - CI failures unrelated.

PR #270, opened 2018-06-26
Add win64 to Pharo VMs

estebanlm:
Working on this PR to add all dependencies to make Pharo run on Windows 64bit.
This is applying what we have talked before (instead building, taking dependencies from mingw versions).
This avoids a lot of problems, but adds a couple of extra libraries to the dependency list... nothing serious or big so we will continue as is for now.


estebanlm:
is working (now I have to adjust things most probably in the 64bit frameworks)

PR #276, opened 2018-08-12
274 primitiveFileStdioHandles() fails to return nil if stdio file is …

akgrant43:
…not available

According to the comment for primitiveFileStdioHandles():

Answer an Array of file handles for standard in, standard out and
standard error, with nil in entries that are unvailable, e.g. because
the platform does not provide standard error, etc.

However, on Windows in practice an array of sqFile entries with null
handles is returned. The image is unable to distinguish between valid
and invalid entries since sqFile is private to the VM.

- Modify sqWin32FilePrims.c() to set the valid mask correctly.
- Modify primitiveFileStdioHandles() to treat a mask of 0 as
  successfully determining that no stdio is available (a normal case
  on Windows).
  A validMask of -1 is treated as an error.
- Update comments to reflect changed behaviour


fniephaus:
Restarting failing builds...

akgrant43:
Thanks, Fabio.

The build got further this time, the newspeak.cog.spur builds failed in .travis_test.sh, but there doesn't seem to be any useful debug information.

Given the main squeak and pharo builds all passed, it seems unlikely that this change is the cause of the newspeak failure.

fniephaus:
All green now!

akgrant43:
Hi Fabio,

Thanks very much for following up on this!

Merging now...

Cheers,
Alistair

fniephaus:
🎉

PR #277, opened 2018-08-12
FileAttributesPlugin 1.3.3: Add path encoding / decoding

akgrant43:
MacOS uses custom decomposed UTF8 encoded strings for path names (while precomposed Unicode strings are typically used within the image).  Encode and decode path names using the existing VM routines (ux2sqPath() and sq2uxPath()).


akgrant43:
MacOS isn't finding a required .h file.  Investigating...

akgrant43:
This has lots of unnecessary changes from trying to figure out OSX behaviour.

Closing and starting again in #280

PR #279, opened 2018-08-19
Suppress rtprio warning when ${CI} env var is set

### LINKED ISSUE #275: Provide control over VM messages
Could the VM provide an option for separating its own output from the image's output?

Concretely, the VM complains about pthread_setschedparam in my Travis builds, which prevents writing command-line scripts with clean output both on stdout and stderr. See https://github.com/hpi-swa/smalltalkCI/issues/388

Maybe it's possible to play with the file descriptors at image startup, but it would make sense if the VM itself had something like a `--silent` or a `--vm-stderr FILE` flag (the JVM has `-Xlog`) to redirect its diagnostic messages so that anything that goes to stdout or stderr exclusively comes from the image.

fniephaus:
This makes the rtprio warning less verbose when the `${CI}` env var is set.
`${CI}` is usually set by common CI providers, see:
- https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
- https://www.appveyor.com/docs/environment-variables/
- https://docs.gitlab.com/ee/ci/variables/#predefined-variables-environment-variables

"pthread_setschedparam failed: Operation not permitted" is still printed to inform about the misconfiguration.

If no one objects, I will merge this in the next couple of days...

Closes #275

### DIFF
--- a/platforms/unix/vm/sqUnixHeartbeat.c
+++ b/platforms/unix/vm/sqUnixHeartbeat.c
@@ -330,6 +330,7 @@ beatStateMachine(void *careLess)
 #else
 # define VMNAME "squeak"
 #endif
+      if (!getenv("CI")) {
         fprintf(stderr, "This VM uses a separate heartbeat thread to update its internal clock\n");
         fprintf(stderr, "and handle events.  For best operation, this thread should run at a\n");
         fprintf(stderr, "higher priority, however the VM was unable to change the priority.  The\n");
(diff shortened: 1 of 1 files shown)


akgrant43:
Hi Fabio,

I think using a CI environment variable to control the output of 
warnings is less than ideal:

- It is doing double duty with a variable, which is never a good idea.

- If I'm running a CI that does allow rtprio and I want the warning, I 
  have to write the non-obvious script to unset the environment 
  variable.

- If I'm running some other production environment and want to suppress 
  the warning I have to do something like:

export CI "this is a ugly hack to make smalltalk suppress one warning"
squeak/pharo some.image

Either of:

--suppress-rtprio-warning  # previously suggested

or

--suppress list,of,warnings

as command line options is a better approach.  We have CI specific 
scripts which can do the appropriate thing.


Cheers,
Alistair

fniephaus:
@akgrant43 I agree, a command line flag is nice and this was just a practical "hack". As I mentioned, the "pthread_setschedparam failed: Operation not permitted" error is still displayed, it's just the "how to fix it" that I tried to suppress here.
I'd like to see that flag implemented soon, but I'm afraid I don't have enough time to work on it (if you know how the option parse work, it should be much work. I just haven't touched that code yet).

akgrant43:
Hi @fniephaus ,

Thanks for the clarification.

Cheers,
Alistair

Issue #288, opened 2018-10-05
Fallback window size on Mac is too small


pavel-krivanek:
Related issue: https://pharo.fogbugz.com/f/cases/20094/Headless-mode-breaks-image

When the image is saved in headless mode, the window size is set to extension 10@10. This value is too small. 

see opensmalltalk-vm/platforms/Mac OS/vm/Developer/sqMacMinimal.c, line 531

krono:
Hey @pavel-krivanek, if you reference the fogbugz site, please at least include the gist of its discussion here. Fogbugz is a closed site and not everybody has an account there.,=

Also, can you tell _why_ `10 @ 10` is too small?
Why does a headless mode bother with the extent in the first place?

pavel-krivanek:
The issue was described with this code
```bash
curl get.pharo.org/60+vm | bash
./pharo Pharo.image save 1.image
./pharo-ui 1.image
```
When you did it, the Pharo started but no UI was shown. Then we realized that the UI actually started and the window was opened but it was so small that it was hard to notice. 
So, when you start the image in headless mode, the VM during saving sets the real window size and since the headless VM has no window, it is set to fallback values (10@10).
Then when the image is started in standard mode with the display, it extremely small.

pavel-krivanek:
I'm pretty sure it is a VM problem. For example on Linux it works as expected (which kind of supports that it is VM related). Mac does not take the display size from the Display but from the window, see the related method:
```c
int ioScreenSize(void) {
	/* return the screen size as two positive 16-bit integers packed into a 32-bit integer */
	int w = 10, h = 10;

	if (stWindow != nil) {
		w = stWindow->portRect.right - stWindow->portRect.left;
		h = stWindow->portRect.bottom - stWindow->portRect.top;
	}
	return (w << 16) | (h & 0xFFFF);  /* w is high 16 bits; h is low 16 bits */
}
```

krono:
(note that `platforms/Mac OS/vm/Developer/sqMacMinimal.c` is no longer used, it is from the deprecated Carbon Mac VM)

krono:
Ok, here's the roundup:

[...]

PR #289, opened 2018-10-05
increse fallback window size on Mac to a standard dimension

pavel-krivanek:
https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/288


krono:
how is  976 by 665 a “standard” dimension? that does not make sense

pavel-krivanek:
just Pharo standard dimension of Pharo window... Squeak has 1024@768. Both are random values from the perspective of this PR.

pavel-krivanek:
well, I updated the PR. I know from where 1024x768 comes from. That's why I wanted a smaller resolution since on displays with XGA resolution is the window too big because of window decorations etc. I meant that from the perspective of this PR the values are "random" magic numbers and it is not a proper solution. But better than nothing.

pavel-krivanek:
Closing this PR. It does not solve the problem in the right place

PR #293, opened 2018-10-06
FileAttributesPlugin 2.0.2: free faPath

akgrant43:
It's too easy to forget that C doesn't have garbage collection,
especially in slang... :-)


akgrant43:
Build failures were attempts to install packages prior to compiling.

PR #295, opened 2018-10-22
Do only handle events that correspond to the VM window

guillep:
While playing with SDL on OSX we have noticed that the main window in the VM is randomly eating events that belong to other windows. This happens because we are consuming events regardless the window they come from.

The corresponding code is in the method pumpRunLoopEventSendAndSignal: in
   https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/platforms/iOS/vm/OSX/sqSqueakOSXApplication%2Bevents.m

I've playing around with making the VM consume only events that come from its own window or from no window (window=null). This seems to work correctly, and I've had run into no issues in the last couple of hours.

I've put a guard so my code is only valid for PHAROVM builds, but I'd gladly remove the guards if people think this is a good move for all the community.

### DIFF
--- a/platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
+++ b/platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
@@ -70,10 +70,28 @@ static int buttonState=0;
 - (void) pumpRunLoopEventSendAndSignal:(BOOL)signal {
     NSEvent *event;
     
+#ifdef PharoVM
+    while ((event = [gDelegateApp.window nextEventMatchingMask:NSEventMaskAny
+                                       untilDate:nil
+                                          inMode:NSEventTrackingRunLoopMode
+                                         dequeue:NO])) {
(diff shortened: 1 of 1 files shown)


eliotmiranda:
Please do remove the guard.  This is clearly the right thing to do on any flavour.  Thanks for this important fix!!

clementbera:
Do you want to review John since I believe you have good OSX experience?

If no one else wants to review I'll merge wednesday afternoon, unless someone has already merged before.

Thanks Guille for this fix. Was very annoying.

johnmci:
Ok I will take a look

johnmci:
Ok it looks ok, not sure if it would affect the older multi-window support but I’m not sure anyone used that anyway.

clementbera:
Thanks John. I was scared because last time someone tried to fix this it had side effect on some events in the main window.

I merge then. We will see if someone complains.

guillep:
I've posted #296 removing the PharoVM guard and adding some comments.

guillep:
BTW Thanks!

PR #296, opened 2018-10-23
Added comments and removed PharoVM guards for windowing support

guillep:
Last integration of #295 went too fast :)

This PR does some cleanups:
 - remove the PharoVM guards,
 - added comments

Answering the questions from the other thread: this should have no effect on systems using a single window. However, with multiple windowing systems,  this makes the consumption of events a cooperative task. If let's say SDL does not consume its events (because we don't call the event processing routine), then those events stay in the queue for ever.

To make it more robust I've been trying to use a separate queue to store alien events and repost them in the queue afterwads, but I could not manage to make it work. And Cocoa's support/documentation does not help here...


eliotmiranda (approved):
I'd rather have this merged and in use than not being tested.  We can fix any breakages later.  As long as the simple case works we can then test the more complex cases more easily bu=y having this in the default VM.

PR #298, opened 2018-10-28
Add minheadless vm

estebanlm:
adds the minheadless sources from @ronsaldo + my own changes. 
this is still WIP (is working for Mac and Linux but not for Win), but to be honest, I want this integrated before the structure reorganisation so I do not need to do it myself *after* ;)


eliotmiranda (commented):
Hi Esteban,

    this looks good, but there is at least one remaining issue.  For example, the changes to platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c eliminate necessary "if (f->isStdioStream) Read/WriteConsole(..." variants.  Can you take a pass over platforms/win32/vm and roll back any incidental changes?

estebanlm:
Hi @eliotmiranda, 

I didn't have time to work on this (and I don't have much time now to be honest, I need to release Pharo7 and there are still too many things open :( )
Specially, because those changes are not mine but from @ronsaldo and I first need to see what he did and what you want to to be able to fix it :P 

I think is to merge latest changes to File plugin there? (no idea how hard would it be :P)

akgrant43:
I can have a look at platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c and platforms/win32/vm if that helps.  And try building the headless version on linux.

ronsaldo:
When I started making the minheadless vm, I rewrote most of the file primitives to force Unicode usage by manually converting from UTF-8 into UTF-16, and removing the usage of tchar for explicit wchar. I remember that I was having Unicode related issues with spanish accents in directories and folders, and also some differences when compiling with MSVC. In parallel, in the main vm branch I think that there was a parallel refactoring on these primitives, so another pass on it should be needed.

akgrant43:
Hi Esteban, Eliot & Ronie,

Ronie, thanks for the history and explanation re Unicode.

Esteban & Ronnie, can I ask what the long term goals are for this code base?

The reason is that it appears to be adding significant complexity to the code base:

It's breaking the existing directory structure, e.g. the two instances of sqUnixCharConv.c are in:

```
./platforms/minheadless/unix/sqUnixCharConv.c
./platforms/unix/vm/sqUnixCharConv.c
```

It adds 69 file names.  Some of these obviously need to be duplicated, e.g. make files are specific to the build type, but other examples include:

- sqUnixCharConv.h is copied unchanged.
- Changes the return types of some functions with the same name, e.g. 'int' to 'sqInt'.
- sqWin32Heartbeat.c where the entire file is copied so that 1 function can be extended (ioHighResClock()).

And with an admission of being too lazy to figure it out, I'm not sure how to build the Windows minheadless variant of, e.g., pharo.cog.spur.  There's:

```
build.linux64x64/pharo.cog.spur.minheadless/build/mvm
```

for linux 64 bit, and the equivalent for linux 32 bit and MacOS 32 bit, but not MacOS 64 bit or Windows.  Should:

```
build.minheadless.cmake/x86/pharo.cog.spur/mvm
```

be used?

I can understand the problems of chasing a moving target, but without a commitment to continue working on this it looks like it will increase technical debt.

Or am I missing something?

Thanks,
Alistair

girba:
I can address the high level concern. I will let the others address the specifics in the code. The high level goal of this code is to enable the VM to not assume a UI, or even a specific UI. This is important for enabling other user interface technologies.

The Windows part does not work yet. However, the goal of this branch is to integrate it now to not lose the work. The Windows support will be added later.

estebanlm:
Medium term, it is intended to use this codebase as the foundation for the official PharoVMs (because we want a real headless option, with the responsibility of starting a world from the image, not the VM).

Of course, that does not means it does not needs work on it. 

Most of the duplicates and that is because when started, @ronsaldo wanted to keep it as separated as possible from the main sources, as a way to preserve them and being able to modify what he needed.

bencoman:
>
>  not sure how to build the Windows minheadless variant of, e.g.,
> pharo.cog.spur.
>

I documented my experience getting the Windows build working with MSVC
here...
http://forum.world.st/Minheadless-trial-td5082569.html

After taking a short break from it I was planning to catch it up to
Esteban's branch
but I hadn't got an opportunity yet.  Seeing Esteban integrating it
prompted me to plan some time
for this at the end of next week, but thats not assured so if you are so
inclined to have a go at it, please do.

From memory this gets at least the StackInterpreter doing simple arithmetic
from the command line,
which should allow you to test file primitives from the command-line.
Esteban mentioned he had
a custom Image for a minheadless+GUI.  I don't remember the state I got Cog
to.

akgrant43:
@girba, @estebanlm & @ronsaldo  thanks for the explanation.  That makes sense, although I think we need to be careful about the duplication.

@bencoman thanks for the write-up, that explains quite a bit that I didn't really understand.

Progress report:

I've successfully built a linux 64 bit cog spur VM and started a standard image with it.  Changes include:

- platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
-- Reinstate stdio processing as highlighted by @eliotmiranda 
-- I also looked at platforms/win32/vm, but the changes there all look fine.
- build.linux64x64/pharo.cog.spur.minheadless/build/mvm
-- Fix the install directory processing and change it so it doesn't write in the middle of the 32 bit VM
- cmake/Plugins.cmake
-- Add FileAttributesPlugin
- third-party/libsdl2.spec
-- fix libSDL2 linux64 so file name

See: https://github.com/akgrant43/opensmalltalk-vm/tree/akg-minheadless-vm

The only issue I've found so far is that right mouse click isn't working.

The MacOS build will still be broken.

Cheers,
Alistair

akgrant43:
Current status:

- The CI build completes for:
-- Windows
-- Main Squeak and Pharo linux and Mac builds
-- Other linux builds
-- Other Mac builds fails in the newspeak tests.  I don't think this is 
   related.

I've only tested linux64 at this stage, but given:

- minheadless are (mostly) separate builds.
- both Esteban and I have done some ad-hoc testing.
- the CI succeeds.
- the fact that this code is more susceptible than most to diverging 
  from the main repository (due to code duplication).
- Esteban and Doru both seem committed to continuing with this work.
- integrating the code now will facilitate others looking at the code, 
  e.g. Ben planned to have a look next week.

it may be worthwhile integrating this now.

What do you think?

Just a reminder, the PR is not this one, but: 

https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/310


My Win32 build fails trying to build and install freetype:


-- Installing: /home/alistair/pharo7/pharo-snap/pharo-vm/opensmalltalk-vm/.thirdparty-cache/windows/i386/include/freetype2/freetype/tttables.h
-- Installing: /home/alistair/pharo7/pharo-snap/pharo-vm/opensmalltalk-vm/.thirdparty-cache/windows/i386/include/freetype2/freetype/tttags.h
-- Installing: /home/alistair/pharo7/pharo-snap/pharo-vm/opensmalltalk-vm/.thirdparty-cache/windows/i386/include/freetype2/ft2build.h
CMake Error at cmake_install.cmake:35 (file):
  file INSTALL cannot find
  "/home/alistair/pharo7/pharo-snap/pharo-vm/opensmalltalk-vm/build.win32x86/pharo.cog.spur/build/third-party/freetype-2.9.1/build/include/freetype/config/ftconfig.h".


I've never used cmake before, so am starting to learn about it, but does 
anyone recognise the above?


Cheers,
Alistair

akgrant43:
This has been added through PR https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/310 which builds on this one.

Closing.

PR #299, opened 2018-10-30
replace valloc() with mmap() when allocating executable pages for cal…

bwestergaard:
…lbacks

Under linux, if SELinux is enabled (i.e. not permisive or disabled), then
the heap memory allocated with vmalloc() and marked as executable with mprotect()
will fail.  In order to have this work with SELinux enabled (and not providing
a module policy file to allow execheap) mmap() should be used.

Note, it might be possible to remove the stdlib.h and sys/mman.h header
file references.

See: http://lists.squeakfoundation.org/pipermail/vm-dev/2018-October/029102.html

### DIFF
--- a/platforms/Cross/plugins/IA32ABI/ia32abicc.c
+++ b/platforms/Cross/plugins/IA32ABI/ia32abicc.c
@@ -292,14 +292,11 @@ allocateExecutablePage(long *size)
 #else
 	long pagesize = getpagesize();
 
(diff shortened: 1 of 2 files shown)


eliotmiranda:
Bob, could you add a comment in front of the mmap call that says this is equivalent to valloc but that it doesn't work on the version of selinux that gave the grief?

PR #311, opened 2018-11-30
Fix the build of the minheadless VM for Pharo in Win32.

ronsaldo:
Quick fixup of the minheadless VM for windows using the CMake building scripts.
The VM builds and works with Cygwin for 32 bits.
It builds with MSVC, but fails with several errors on runtime.


bencoman:
Hi Ronie,

What is your procedure for building with MSVC ?
I'd like to have a go at it.

cheers -ben

PR #313, opened 2018-12-01
Fixes for 64-bit MSVC minheadless build

bencoman:
Following on from @ronsaldo's PR #311, these are some additional fixes for 64-bit MSVC build from my minheadless trial (http://forum.world.st/Minheadless-trial-td5082569.html).  We both made some changes around 32/64 bit detection in CMakeLists.txt so that still needs some work. All my other fixes were already covered in PR #311.

### DIFF
--- a/platforms/Cross/plugins/IA32ABI/x64win64abicc.c
+++ b/platforms/Cross/plugins/IA32ABI/x64win64abicc.c
@@ -38,6 +38,9 @@ extern
 #endif
 struct VirtualMachine* interpreterProxy;
 
+#ifdef _MSC_VER
+# define alloca _alloca
+#endif
 #if __GNUC__
 # define setsp(sp) __asm__ volatile ("movq %0,%%rsp" : : "m"(sp))
 # define getsp() ({ void *sp; __asm__ volatile ("movq %%rsp,%0" : "=r"(sp) : ); sp;})
--- a/platforms/minheadless/windows/sqPlatformSpecific-Win32.c
+++ b/platforms/minheadless/windows/sqPlatformSpecific-Win32.c
@@ -48,11 +48,27 @@
 
 # define fopen_for_append(filename) fopen(filename,"a+t")
 
-/* default fpu control word:
-   _RC_NEAR: round to nearest
-   _PC_53 :  double precision arithmetic (instead of extended)
-   _EM_XXX: silent operations (no signals please)
+/* default fpu control word and mask: 
+   _MCW_RC: Rounding control
+       _RC_NEAR: round to nearest
+   _MCW_PC: Precision control
+       _PC_53:  double precision arithmetic (instead of extended)
+   _MCW_EM: Interupt exception mask
+       _EM_XXX: silent operations (no signals please)
+   https://docs.microsoft.com/en-us/previous-versions/e9b52ceh%28v%3dvs.140%29
 */
+
+#if !defined(_MCW_PC) 
+// x64 does not support _MCW_PC
+// The x64 CPU hardware default is 64-bit precision mode (80-bit long double); 
+// Microsoft expects software to set 53-bit mode before any user mode x87 instructions 
+// are reached. Microsoft made a change in responsibility for initializing precision mode 
+// in the X64 OS. The X64 OS sets 53-bit mode prior to starting your .exe, where the 32-bit OS 
+// expected the program to make that initialization.
+# define _MCW_PC 0
+#endif
+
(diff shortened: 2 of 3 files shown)


eliotmiranda:
Hi Ben,

platforms/minheadless/windows/sqPlatformSpecific-Win32.c still has to cover both x64 and x86 so IMO the right way to write this:

```
    /* Setup the FPU */
    // x64 does not support _MCW_PC or _MCW_IC (https://msdn.microsoft.com/en-us/library/e9b52ceh.aspx)
     _controlfp(FPU_DEFAULT, _MCW_EM | _MCW_RC );

```
is

```
    /* Setup the FPU */
    // x64 does not support _MCW_PC or _MCW_IC (https://msdn.microsoft.com/en-us/library/e9b52ceh.aspx)
#if !defined(_MCW_PC)
# define _MCW_PC 0
#endif
#if !defined(_MCW_IC)
# define _MCW_IC 0
#endif 
     _controlfp(FPU_DEFAULT, _MCW_EM | _MCW_RC | _MCW_PC | _MCW_IC);
```

eliotmiranda:
Hi Ben,

   put the definitions at the top of the file or at the first occurrence of _controlfp as you see fit.  It does look as if we can ignore _MCW_IC.  _MCW_PC on the other hand is vital.  If the VM called foreign code though the FFI and that foreign code changed the _MCW_PC mode before calling back, unless the callback entry code resets the fp mode back to the default, the VM's floating point results would be incorrect.  Essentially callbacks must reinitialize the processor so that it functions as expected on each callback in case foreign code puts the processor into some other mode.  With simple processors this isn't an issue but with the x86 it is.

An alternative may be to jettison the entire use of _controlfp if we can be convinced that the VM only makes use of SSE instructions to implement floating point.  That's the case with the JIT floating point code.  And we use -SSE3 flags on all compilers I'm aware of.  If we go this route I would still include the _controlfp code, surrounded by #if 0 ... #endif, simply to document the issue.

bencoman:
Hi Eliot, I think I captured everything discussed.  

One extra thing though...
I was wondering that if x64 doesn't support _MCW_RC, then the _PC_53 is irrelevant, thus our 64-bit VM would be using the CPU default 64-bit extended floating point and be different from our 32-bit VM, but then I got [this insight](https://software.intel.com/en-us/forums/software-tuning-performance-optimization-platform-monitoring/topic/277814)...
> The CPU hardware default is 64-bit precision mode (80-bit long double); Microsoft expects software to set 53-bit mode before any user mode x87 instructions are reached. Microsoft made a change in responsibility for initializing precision mode in the X64 OS. The X64 OS sets 53-bit mode prior to starting your .exe, where the 32-bit OS expected the program to make that initialization.

If that is how you understand it, I think that explanation could be a useful code comment.

eliotmiranda:
Hi Ben,

    ideally we'd have bit-identical FP on all platforms.  Currently this is only achievable using Andreas Raab's FloatMath plugin, used by Croquet and Terf, and not yet tested in 64-bits.  What that has to say about whether we support the default or 53-bit on x64 I don't know.  i guess we should for now simply support the default m ode.  In any case it's unlikely to make much difference because the computational model yields results being truncated to 53 bits after each operation (i.e. other than in the graphics subsystem, floating point operations are performed on double-precision floats so there's no scope for extended calculations to use 64-bit precision internally.  So feel free to include the above as commentary but I think we want to apply your pull request asap and so we should leave the precision argument for a later day when we have more data (such as a 64-bit Terf to test).

eliotmiranda:
Hi Nicolas,  thanks for the reminder.  The problem is indeed that that plugin is written w.r.t. fdlibm which is effectively obsolete.  However there is an alternative.  See first this for a good overview of the issue:
http://christian-seiler.de/projekte/fpmath/
and then this for a widely used solution:
https://www.mpfr.org
This still presents problems; we'd hav e to modify the JIT to generate analogous code to Gnu's MPFR.  But at least we have a reference implementation against which to test.
We should perhaps take this discussion to a new issue.

bencoman:
Okay, its ready to go. Please merge.

nicolas-cellier-aka-nice:
mpfr is multiple precision (arbitrary precision if you prefer). IMO it's too much for FloatMathPlugin.
CRLIBM (correctly rounded mathematical libray) could be a good modern alternative to fdlibm (from INRIA too).

CRLIBM is also LGPL
https://hal-ens-lyon.archives-ouvertes.fr/ensl-01529804/file/crlibm.pdf

PR #318, opened 2018-12-06
Fix compile for Windows serial

bencoman:
Recently platforms/Cross/plugins/SerialPlugin/sqNullSerialPort.c 
was copied from platforms/iOS/plugins/SerialPlugin/sqNullSerialPort.c
but "noErr" enum from https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-18.1/MacTypes.h
doesn't exist on Windows. 
Change it to a #define.


bencoman:
CANCEL: Eliot solved it another way.

PR #326, opened 2018-12-21
Check sanity of curled gzip data

shingarov:
The getGoodSpur*VM.sh scripts assume that the VM mentioned in the
topmost notification, is always there on bintray.  This is not
always the case.  Unfortunately, bintray will not return a 404 code
on a 404 error, but instead give a 200 with garbage data.  While we
can't do anything about that, at least make an attempt to bail if
the file doesn't even look like gzip-compressed data.

### DIFF
--- a/image/getGoodSpurVM.sh
+++ b/image/getGoodSpurVM.sh
@@ -52,6 +52,14 @@ else
 				exit
 			fi
 			curl -L "$URL" -o "$LATESTVM"
+			# bug in bintray: on a 404 error,
+			# it will return a 200 with garbage data
+			if [[ $(file "$LATESTVM" | grep 'gzip compressed data') ]]; then 
+			  tar xzf "$LATESTVM"
+			else
+			  echo No gzip data at "$URL"
+			  exit
+			fi
 			tar xzf "$LATESTVM"
 			mv sqcogspurlinuxht $VM
 			rm -f "$LATESTVM"
--- a/image/getGoodSpur64VM.sh
+++ b/image/getGoodSpur64VM.sh
@@ -52,7 +52,14 @@ else
 				exit
 			fi
 			curl -L "$URL" -o "$LATESTVM"
-			tar xzf "$LATESTVM"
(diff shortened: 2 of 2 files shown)


akgrant43:
This looks good, however shouldn't it "exit 1" on error so the calling script knows not to proceed?

Bonus points :-): It would be really nice if it could retrieve the previous version (but I haven't looked at bintray to see if they is practical).

image/getGoodSpurVM.sh
```diff
@@ -52,6 +52,14 @@ else
...
 			curl -L "$URL" -o "$LATESTVM"
+			# bug in bintray: on a 404 error,
+			# it will return a 200 with garbage data
+			if [[ $(file "$LATESTVM" | grep 'gzip compressed data') ]]; then 
+			  tar xzf "$LATESTVM"
+			else
+			  echo No gzip data at "$URL"
+			  exit
+			fi
 			tar xzf "$LATESTVM"
```
gcotelli: This line seems to be duplicated. It's already contained in the if

akgrant43:
Looks good.  I'll leave it a bit longer in case there is any more feedback and then merge it in.

Thanks,
Alistair

eliotmiranda:
Hi Boris,  I’m wondering if the right thing to do is to build a vm that includes the processor simulator plugins in the relevant ../build.* directories.  The makefiles are set up to build the processor simulator plugins only if the support libraries (the processor simulator code itself) have been built, so by default the CI does not build them.

  I would prefer that we set up a CI to build the processor simulator plugins and download and install these as part of the getGoodSpur*VM scripts, but that’s work I haven’t found time for and at least on Mac installing a plugin in the .app will break the signature and provoke a warning on launch.  So building a vm in place, while slow and not yet automated might be the right approach.  It does mean running the conf.COG scripts and make in the relevant build.*/{bochsx86,bochsx64,gdbarm32} directories, none of which is as smooth as it might be.  What do you think?

PR #329, opened 2018-12-28
Win64 cleanups

nicolas-cellier-aka-nice:
This is essentially sanitizing and cleanups of existing win32 platform code.
There's no new feature or behavior change except this one: path passed to security plugin will be interpreted as UTF-8 encoded on windows VM.
These changes are required before cleaning the contract between image and VM, especially about encoding of various strings passed: they should be encoded in UTF8 which seems best for platform interoperability.
It's a step forward for enabling the -DUNICODE as attempted by Tobias about 2 years ago.

It is also very important to reduce the number of compiler warnings. 9 times out of 10, these warnings are false positive, but the tenth is indicating a true problem that we'd better not ignore. If we let the benign warnings overflow us, then we are depriving ourself of a useful tool.

### DIFF
--- a/platforms/win32/plugins/CroquetPlugin/sqWin32CroquetPlugin.c
+++ b/platforms/win32/plugins/CroquetPlugin/sqWin32CroquetPlugin.c
@@ -8,7 +8,7 @@ static BOOLEAN (__stdcall *RtlGenRandom)(PVOID, ULONG) = NULL;
 int ioGatherEntropy(char *bufPtr, int bufSize) {
   if(!loaded) {
     loaded = 1;
-    hAdvApi32 = LoadLibrary("advapi32.dll");
+    hAdvApi32 = LoadLibraryA("advapi32.dll");
     RtlGenRandom = (void*)GetProcAddress(hAdvApi32, "SystemFunction036");
   }
   if(!RtlGenRandom) return 0;
--- a/platforms/win32/vm/sqWin32.h
+++ b/platforms/win32/vm/sqWin32.h
@@ -263,7 +263,7 @@ extern char imageName[];		/* full path and name to image */
 extern TCHAR imagePath[];		/* full path to image */
 extern TCHAR vmPath[];		    /* full path to interpreter's directory */
 extern TCHAR vmName[];		    /* name of the interpreter's executable */
-extern TCHAR windowTitle[];             /* window title string */
+extern char windowTitle[];             /* window title string */
 extern char vmBuildString[];            /* the vm build string */
 extern TCHAR windowClassName[];    /* class name for the window */
 
--- a/platforms/win32/vm/sqWin32Backtrace.c
+++ b/platforms/win32/vm/sqWin32Backtrace.c
@@ -261,14 +261,14 @@ get_modules(void)
 {
 	DWORD moduleCount2, i;
 	HANDLE me = GetCurrentProcess();
-	HANDLE hPsApi = LoadLibrary("psapi.dll");
+	HANDLE hPsApi = LoadLibraryA("psapi.dll");
 	HMODULE *modules;
 
 	EnumProcessModules = (void*)GetProcAddress(hPsApi, "EnumProcessModules");
 	GetModuleInformation=(void*)GetProcAddress(hPsApi, "GetModuleInformation");
 
 	if (!EnumProcessModules(me, (HMODULE *)&modules, sizeof(modules), &moduleCount)) {
-		printLastError("EnumProcessModules 1");
+		printLastError(TEXT("EnumProcessModules 1"));
 		return;
 	}
 	modules = malloc(moduleCount);
@@ -280,14 +280,14 @@ get_modules(void)
 	all_exports = calloc(moduleCount / sizeof(HMODULE) + EXTRAMODULES,
 						 sizeof(dll_exports));
 	if (!modules || !all_exports) {
-		printLastError("get_modules out of memory");
+		printLastError(TEXT("get_modules out of memory"));
 		if (modules)
 			free(modules);
 		return;
 	}
 
 	if (!EnumProcessModules(me, modules, moduleCount, &moduleCount2)) {
-		printLastError("EnumProcessModules 2");
+		printLastError(TEXT("EnumProcessModules 2"));
 		free(modules);
 		return;
 	}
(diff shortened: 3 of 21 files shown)


platforms/win32/plugins/CroquetPlugin/sqWin32CroquetPlugin.c
```diff
@@ -8,7 +8,7 @@ static BOOLEAN (__stdcall *RtlGenRandom)(PVOID, ULONG) = NULL;
 int ioGatherEntropy(char *bufPtr, int bufSize) {
   if(!loaded) {
     loaded = 1;
-    hAdvApi32 = LoadLibrary("advapi32.dll");
+    hAdvApi32 = LoadLibraryA("advapi32.dll");
```
krono, line 11: What about

```suggestion
    hAdvApi32 = LoadLibrary(TEXT("advapi32.dll"));
```

instead?

krono:
I really like this PR.

There's one more "fundamental" comment tho:
I'd think that we should not change Windows code to contain any _explicit_ `SomFunc…A(…)` calls any more.
I'd rather say, 
 1. Either do a conversation,
 2. Use a `TCHAR` variant or,
 3. Explicitly use the `SomeFunc…W(…)` variant.

PS: I just read the last Windows version _not_ being UNICODE was Windows ME. Everything after that internally is `…W` in the first place…

platforms/win32/vm/sqWin32.h
```diff
@@ -263,7 +263,7 @@ extern char imageName[];		/* full path and name to image */
 extern TCHAR imagePath[];		/* full path to image */
 extern TCHAR vmPath[];		    /* full path to interpreter's directory */
 extern TCHAR vmName[];		    /* name of the interpreter's executable */
-extern TCHAR windowTitle[];             /* window title string */
+extern char windowTitle[];             /* window title string */
```
krono, line 266: Why?

platforms/win32/vm/sqWin32Backtrace.c
```diff
@@ -261,14 +261,14 @@ get_modules(void)
 {
 	DWORD moduleCount2, i;
 	HANDLE me = GetCurrentProcess();
-	HANDLE hPsApi = LoadLibrary("psapi.dll");
+	HANDLE hPsApi = LoadLibraryA("psapi.dll");
```
krono, line 264: See above

platforms/win32/vm/sqWin32Heartbeat.c
```diff
@@ -65,6 +65,11 @@ sqLong ioHighResClock(void) {
 						: "=a" (value)
 						: 
 						: "rdx");
+#elif defined(_MSC_VER)
```
krono, line 68: 👍

platforms/win32/vm/sqWin32Main.c
```diff
@@ -540,7 +540,7 @@ void gatherSystemInfo(void) {
 
 
   {
-    HANDLE hUser = LoadLibrary( "user32.dll" );
+    HANDLE hUser = LoadLibraryA( "user32.dll" );
```
krono, line 543: see above

platforms/win32/vm/sqWin32Prefs.c
```diff
@@ -207,8 +207,17 @@ void LoadPreferences()
   }
 
   /* get window title from ini file */
+#ifdef UNICODE
+  {
+    TCHAR windowTitleT[MAX_PATH];
+    GetPrivateProfileString(U_GLOBAL, TEXT("WindowTitle"),
+      TEXT(""), windowTitleT, MAX_PATH, squeakIniName);
+    WideCharToMultiByte(CP_UTF8, 0, windowTitleT, -1, windowTitle, MAX_PATH, NULL, NULL);
```
krono: This needs a check

nicolas-cellier-aka-nice:
Hi Tobias, about LoadLibraryA, I see nothing in MSDN suggesting that these ASCII versions will be deprecated.
https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-loadlibrarya

The purpose of using W variants is to enable usage of internationalized strings, like in window titles, file names, etc... But when we have a simple explicit ASCII string like "advapi32.dll", using the W variant brings nothing.

About windowTitle, I can explain more: if you look at https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/476f70605a0352dd7528d251f7403e9233716cdb/platforms/win32/vm/sqWin32Window.c

you'll see this: `if(*windowTitle) sprintf(titleString, "%s", windowTitle);` we first copy the `char *windowTitle` to the `char *titleString`. It obviously cannot be UTF16 at this stage, otherwise the copy would be shorten at first ASCII character.

Then `titleString` is interpreted as UTF8, converted to UTF16 before being displayed as title, line 773:

    MultiByteToWideChar(CP_UTF8, 0, titleString, -1, wideTitle, MAX_PATH+20);
    SetWindowTextW(stWindow, wideTitle);

This was before I change anything. Also we have this function:

    char *ioGetWindowLabel(void) {
      return windowTitle;
    }

I think it is for the purpose of communicating the title back to the image. For communicating with the image, we wish using UTF8, so the minimal and simplest change was to just declare `windowTitle` as char * (which it is now because we can't use -DUNICODE yet), and align the squeak.ini preference to read in UTF16 and convert back to UTF8. You are right, I messed up this part (because I currently compile with `-DUNICODE` for the purpose of discovering the flaws, when I should rather compile with AND without this option). I will issue another commit and use `GetPrivateProfileStringW` unconditionnally, which is a lot simpler.

In fact, most of the time, we have to communicate the strings with the image, and for that, we will need to perform UTF8-UTF16 conversions, so compiling with `-DUNICODE` is not enough and not really a goal per se. Using the generic versions and generic TCHAR *string has only a value if we want to maintain both an ASCII version and a Wide version. But as you said, the ASCII version does not have much interest now, so we could as well directly use W version where we think we need them, or for making the code a bit more uniform (polishing). If it adds un-necessary complexity, I would say just stick to the A version.

platforms/win32/vm/sqWin32Prefs.c
nicolas-cellier-aka-nice: Definitely! this can only work with -DUNICODE... I will emit a new commit.

nicolas-cellier-aka-nice: Wait, I have been abused by the too-short context: this piece of code is precisely protected with a `#ifdef UNICODE`, so finally I think it is correct.

We may switch to using the W version unconditionnally, but for that, `squeakIniName` must also be UTF-16 encoded. This is part of larger changes that I have cooking and did not commit yet.

krono: hm, I just meant that `WideCharToMultiByte` needs a success check :)

krono:
I agree with most of what you said.
I'd rather have consistency when "conversing" with the Windows API. I.e. Either TCHAR, Or …A Or …W, but if possible not a mixture, to also in the future understand what we were doing here :)

Apart from that, good work!

platforms/win32/vm/sqWin32Prefs.c
nicolas-cellier-aka-nice: OK!

PR #330, opened 2018-12-30
FileAttributesPlugin 2.0.8

akgrant43:
Get the plugin simulator working again.

Until the first call to #primitiveFailForOSError:, at which point the
simulator fails (the real plugin works). Still on the ToDo list...


nicolas-cellier-aka-nice:
No need to wait full completion of CI, enough versions are already working

PR #334, opened 2019-01-01
[IGNORE]  Dummy commit to allow test CI build from known base

akgrant43:


akgrant43:
This is annoying... the build fails in a different plugin every time.

nicolas-cellier-aka-nice:
Maybe you should remove -j4 like Ben did.

akgrant43:
> Maybe you should remove -j4 like Ben did.

Thanks!

> So after further observation of build times, maybe -j4 should be removed permanently.
> Clearly understandable log files seems more important.

+1

nicolas-cellier-aka-nice:
Look:
https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/src/plugins/SqueakFFIPrims/SqueakFFIPrims.c removed the `#if defined(__arm__)` so as to enable __arm64__

From my understanding, this is the only macro that enables recognizing arm on gcc
http://infocenter.arm.com/help/topic/com.arm.doc.dui0491i/BABJFEFG.html

akgrant43:
Defining ARM32 allows the build to succeed.  It will be tonight before I get a chance to tidy this up and try a build for real.

nicolas-cellier-aka-nice:
Fixed in https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/1c6552daac9cc823b3a74b643676b3a346f0bf5d

Note: at least, all travis builds are green
I don't wait for appveyor status because the queue is too deep, but we did not change anything related to windows in latest commits.

akgrant43:
Hi Nicolas & Ben,

Thanks for your assistance with tracking this down and fixing it.

Just FYI: I fired up a Raspberry Pi I have lying around and checked that the new VM will at least start and can run LibC>>system:.

Cheers,
Alistair

PR #343, opened 2019-01-04
openssl: Update from 1.0.2l to 1.0.2q

akgrant43:
We should be using the latest openssl library to avoid known vulnerabilities.


akgrant43:
Recreating without the noise.

PR #349, opened 2019-01-05
.appveyor.yml test - try to add [skip appveyor]

akgrant43:
Based on https://github.com/appveyor/ci/issues/1865

[skip appveyor]

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -97,6 +97,11 @@ cache:
   - .thirdparty-cache
 
 install:
+  # Search for [appveyor skip], [skip appveyor], [skip ci] or [ci skip] and exit if found in full commit message
+  - ps: $commit=$env:APPVEYOR_REPO_COMMIT_MESSAGE + $env:APPVEYOR_REPO_COMMIT_MESSAGE_EXTENDED
+  - ps: If ($commit.Contains("[skip appveyor]") -Or $commit.Contains("[appveyor skip]")) { Exit-AppVeyorBuild }
+  - ps: $commit=$env:APPVEYOR_REPO_COMMIT_MESSAGE + $env:APPVEYOR_REPO_COMMIT_MESSAGE_EXTENDED
+  - ps: If ($commit.Contains("[skip ci]") -Or $commit.Contains("[ci skip]")) { Exit-AppVeyorBuild }
   - 'curl -fsSL --retry 4 -m 600 -o "setup-x86.exe" "http://cygwin.com/setup-x86.exe" '
   - 'curl -fsSL --retry 4 -m 600 -o "setup-x86_64.exe" "http://cygwin.com/setup-x86_64.exe" '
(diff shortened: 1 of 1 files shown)


fniephaus:
Hi @akgrant43,
Have you seen this? https://www.appveyor.com/docs/how-to/filtering-commits/

akgrant43:
Hi Fabio,

My understanding with the default commit filter is that it requires the [skip ci] message in the commit title.  Most people don't want the first line of the commit polluted by this type of message, so the idea here is to allow the [skip ci] to be placed anywhere in the commit message (it has it's own drawbacks, see https://help.appveyor.com/discussions/problems/11603-skip-ci-doesnt-work-as-expected for a discussion (thanks to Ben for providing this link).

Cheers,
Alistair

nicolas-cellier-aka-nice:
Interesting... We could be using something like this

    only_commits:
        files:
            - *src/
            - platforms/win32/
            - platforms/Cross/
            - build.win*/
            - third-party/

akgrant43:
Hi Nicolas,

> Interesting... We could be using something like this
> 
> ```
> only_commits:
>     files:
>         - *src/
>         - platforms/win32/
>         - platforms/Cross/
>         - build.win*/
>         - third-party/
> ```

The drawback of this approach is that we identify the VM by the commit id, and this will prevent the same commit id being present on all platforms whenever a change is made that doesn't affect Windows.

fniephaus:
>The drawback of this approach is that we identify the VM by the commit id, and this will prevent the same commit id being present on all platforms whenever a change is made that doesn't affect Windows.

Well shouldn't it work that way? I mean, there are other drawbacks of identifying VMs by commit id (e.g. the need for git hooks). More importantly, why rebuild all Linux/macOS VMs when only Windows code was modified? I have never expected bleeding edge "releases" on Bintray to be complete, no one really should because they are mainly for development and builds happen to fail from time to time. Stable release builds will still be triggered simultaneously on AppVeyor and Travis. So we should be fine, right?

akgrant43:
> Well shouldn't it work that way? I mean, there are other drawbacks of identifying VMs by commit id (e.g. the need for git hooks).

I'm not sure.  If we introduced a version number or build number there may still be a need to substitute it in to other files, so there may still be a use for the commit hooks.  And event with the other identifier, I wouldn't want to get rid of the CommitHash - it makes it easy as a developer to find the code.


> More importantly, why rebuild all Linux/macOS VMs when only Windows code was modified?

This is a good point, of course.


> I have never expected bleeding edge "releases" on Bintray to be complete, no one really should because they are mainly for development and builds happen to fail from time to time. Stable release builds will still be triggered simultaneously on AppVeyor and Travis. So we should be fine, right?

Implicit in what you're saying is that someone then manually starts a full build for every release candidate.

So, I'm fine with the change you're proposing, it just requires buy-in from the people that are actually responsible for the releases in the various flavours (and thus are responsible for ensuring a full build for the released VM).

It also doesn't completely remove the usefulness of being able to skip building when desired. :-)

PR #351, opened 2019-01-08
Simplify gitignore build directories [skip ci]

bencoman:
I noticed that modified Makefiles weren't being pickup up by `git status`.
It was too tedious trying to fix the existing definitions, so I've condensed them down to 4 lines from 34.
I'm reasonably sure the result is the same, but if I've missed something its low impact, so maybe just integrate first to let everyone suck-n-see.  
Doing a `git status` before each `git commit` will let you confirm that its working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,40 +22,15 @@
 platforms/unix/config/autom4te.cache/
 
 # Don't track build directories 
-/build*/**/*
-# except do track these files
-!branding.gmk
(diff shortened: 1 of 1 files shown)


bencoman:
Bah! HOLD for a minute

krono:
@bencoman Are you interested in reviewing thas since the `.gitignore` changed?

krono:
closing for now

PR #353, opened 2019-01-16
Patch callback thunkEntry to not optimize

guillep:
O2 and O1 compilation produce a segmentation fault due to stack corruption (when debugging we saw some extra pushes from the stack) on callbacks return. This happens on windows 32, when compiling with mingw gcc 7.4.0.

The issue can be reproduced easily by running the Alien qsort example in latest vms in both Pharo and Squeak.

This PR proposes to patch just the thunkEntry function. Not optimizing just that function solves the issue in our environment, though maybe there is a more fine-grained solution. We should still investigate what is the particular optimization that causes the problem.

### DIFF
--- a/platforms/Cross/plugins/IA32ABI/ia32abicc.c
+++ b/platforms/Cross/plugins/IA32ABI/ia32abicc.c
@@ -169,7 +169,7 @@ getMostRecentCallbackContext() { return mostRecentCallbackContext; }
  * args would get copied into a struct on the stack. A pointer to the struct
  * is then passed as an element of the VMCallbackContext.
  */
-long
+long __attribute__((optimize("O0")))
 thunkEntry(void *thunkp, sqIntptr_t *stackp)
 {
 	VMCallbackContext vmcc;


nicolas-cellier-aka-nice:
Sounds good.
Note that `Alien exampleCqsort` works when VM is compiled with `mvm -f` and clang...
Just change the default compiler in https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/build.win32x86/common/Makefile.tools

I'm using this version which seems up-to-date on cygwin distrib:

> $ i686-w64-mingw32-clang --version
> clang version 5.0.1 (tags/RELEASE_501/final)
> Target: i686-w64-windows-gnu
> Thread model: posix
> InstalledDir: /usr/bin

marceltaeumel:
qsort still SEGFAULTs for me when compiled via Cygwin with `gcc-Version 10.2.0 (GCC)` in production mode (`./mvm -f`). :-( 

When trying to qsort a list of 10 floats, the first callback is fine but the second callback has a different address for the callback context -- which should not happen -- and shows bogus values for the stack arguments. Maybe it is related to some sort of stack optimization?

Works fine with DEBUG build, does also SEGFAULT with ASSERT build.

marceltaeumel:
I think I will force the optimization to `-O0` in `ia32abicc.c` since we still do not understand why `thunkEntry` gets messed up like this in 32-bit builds using `gcc`. But keep it optimized for `clang`.

```
#if defined(__GNUC__) && !defined(__clang__)
__attribute__((optimize("O0")))
#endif
```

PR #354, opened 2019-01-17
Ignore missing dylibs

guillep:
The ignore mechanism was only implemented for bundles, but not for dylibs, used only in Pharo builds.
This PR introduces the same mechanism as in bundles for dylibs, both in 32 and 64 bit builds.

This should fix pharo osx builds, with no effects on other platforms or builds.


krono:
You replaced tabs by spaces. That's not gonna work correctly in a Makefile…

guillep:
Yes, I realized that ^^

The latest build is green :)

PR #356, opened 2019-01-21
arm64ffi progress

KenDickey:
Should have 3 unit test failures..


eliotmiranda:
Looks good.  ARM64FFIPlugin.c needs to be regenerated with a clean VMMaker.oscog to avoid having a dirty package.  But we can fix that soon enough.  Ken, you're going to commit those changes to VMMakerInbox right?  Once done I can regenerate ARM64FFIPlugin.c

eliotmiranda:
oops :-(  didn't mean to close.  Sorry for the noise.

PR #364, opened 2019-01-29
pharo: Forward the packaging directory from pharo-vm here

zecke:
Add structure and rules to build a source package for Debian/Ubuntu
and use it for RPM based distributions. A source package can be built
by the Open Build Service (OBS) or by users. It is intended to be
rebuildable by any user.

Next steps are to hook this into the travis-ci build to upload a
new source package on every build into an experimental feed.


bencoman:
Great initiative. Thanks Holger. 
Just curious what the concept is behind having a new top-level directory "packaging" rather than a directory under "deploy/pharo" ?

fniephaus:
+1 for moving this to `deploy/pharo`.

PR #366, opened 2019-02-02
365: Remove PharoV50.sources from linux builds

akgrant43:
Pharo linux builds are still including PharoV50.sources, which isn't
used in any currently supported versions.  For Pharo7 and later sources
are generated for each build, so there's no need to ever include
sources.


fniephaus (approved):
+1 for removing obsolete code.

PR #367, opened 2019-02-02
Move the directory into deploy/

zecke:
No need for another top-level directory. I didn't move it into
deploy/pharo/ as we might want to use some of the same files for
squeak.


fniephaus:
Thanks! +1 for using all/some of this for Squeak.

PR #373, opened 2019-02-28
fix event memory leak

maenu:
Fixes a memory leak in the macOS event loop, see discussion on mailing list http://lists.squeakfoundation.org/pipermail/vm-dev/2019-February/030595.html

### DIFF
--- a/platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
+++ b/platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
@@ -73,6 +73,8 @@ static int buttonState=0;
 // If the event does not correspond to this window, we take it from the event queue anyways and re-post it afterwards
 // This gives other windows the opportunity to consume their events
 - (void) pumpRunLoopEventSendAndSignal:(BOOL)signal {
+
+	@autoreleasepool {
        NSEvent *event;
        NSMutableArray *alienEventQueue = [[NSMutableArray alloc] init];
        while ((event = [NSApp nextEventMatchingMask:NSEventMaskAny
@@ -89,14 +91,14 @@ static int buttonState=0;
                interpreterProxy->signalSemaphoreWithIndex(gDelegateApp.squeakApplication.inputSemaphoreIndex);
            }
          }
-     }
-
-     // Put back in the queue all events that did not belong to this window
-     // They will be managed by other windowing systems
-     // (or by a subsequent invocation to this function)
-     while ((event = [alienEventQueue firstObject])) {
-       [NSApp postEvent: event atStart: NO];
-       [alienEventQueue removeObject: event];
+       }
+       // Put back in the queue all events that did not belong to this window
+       // They will be managed by other windowing systems
+       // (or by a subsequent invocation to this function)
+       while ((event = [alienEventQueue firstObject])) {
+         [NSApp postEvent: event atStart: NO];
+         [alienEventQueue removeObject: event];
+       }
      }
 }
 


platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
```diff
@@ -73,6 +73,8 @@ @implementation sqSqueakOSXApplication (events)
 // If the event does not correspond to this window, we take it from the event queue anyways and re-post it afterwards
 // This gives other windows the opportunity to consume their events
 - (void) pumpRunLoopEventSendAndSignal:(BOOL)signal {
+
+	@autoreleasepool {
        NSEvent *event;
        NSMutableArray *alienEventQueue = [[NSMutableArray alloc] init];
```
krono, line 79: @johnmci tries to make sure that everything builds on ARC and non-ARC, so maybe do this:
```suggestion
       NSMutableArray *alienEventQueue = AUTORELEASEOBJ([[NSMutableArray alloc] init]);
```

platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
```diff
@@ -89,14 +91,14 @@ - (void) pumpRunLoopEventSendAndSignal:(BOOL)signal {
...
-       [NSApp postEvent: event atStart: NO];
-       [alienEventQueue removeObject: event];
+       }
+       // Put back in the queue all events that did not belong to this window
+       // They will be managed by other windowing systems
+       // (or by a subsequent invocation to this function)
+       while ((event = [alienEventQueue firstObject])) {
+         [NSApp postEvent: event atStart: NO];
+         [alienEventQueue removeObject: event];
+       }
```
krono, line 101: …and this
```suggestion
       }
       RELEASEOBJ(alienEventQueue);
```

maenu:
I used `@autoreleasepool` because it is used elsewhere: https://github.com/OpenSmalltalk/opensmalltalk-vm/search?q=autoreleasepool&unscoped_q=autoreleasepool

Having said that, I have virtually no experience with Objective-C and on the coding conventions for the VM, so expert opinions are highly appreciated. Also, we need to make sure that the solution does not have unwanted side-effects. If I am understanding the code correctly, we only have to make sure that `alienEventQueue` is eventually released.

krono:
the `@autoreleasepool` is ok, leave it there. Just maybe add the other things too :)

maenu:
I do not fully get the idea of `AUTORELEASEOBJ` and friends, this needs to be tested again. Guille mentioned that `[alienEventQueue release]` lead to segfaults: http://lists.squeakfoundation.org/pipermail/vm-dev/2019-February/030607.html. Not sure if this might cause issues.

krono:
Yea, don't `release` manually.
The macros get populated depending on availability of ARC: https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/platforms/iOS/vm/SqueakPureObjc_Prefix.pch

So, when ARC is present (it mostly is nowadays), both macros do exaclty nothing.

maenu:
Hm, where `@autoreleasepool` is used, `RELEASEOBJ` is mostly not used. Isn't `@autoreleasepool` enough, even for non-ARC?

krono:
Don't know. That's why I tagged @johnmci

johnmci (changes requested):
Ok, as @krono pointed out 
NSMutableArray *alienEventQueue = AUTORELEASEOBJ([[NSMutableArray alloc] init]);
with the autoreleasepool. 

I'm not sure anything is build without ARC anymore, but it should be there to be correct for people who need to build on older platforms. 

AUTORELEASEOBJ with ARC does nothing, and the autoreleasepool wrapper then is freeing autorelease objects in the scope of this routine. Since it is pumped from the VM processing requests there isn't a wrapping autorelease pool higher up in the stack.  Easy test is to do invoke the "About Menu"

krono:
> Ok, as @krono pointed out
> NSMutableArray *alienEventQueue = AUTORELEASEOBJ([[NSMutableArray alloc] init]);
> with the autoreleasepool.

Do we need the `RELEASEOBJ` in this case?

krono:
John, thank you for clearing up what's going on

BTW: could you open an issue for the autoreleasepool-on-callout?

platforms/iOS/vm/OSX/sqSqueakOSXApplication+events.m
krono: Ok, this no.

johnmci:
I will be at Camp Smalltalk at end of March, likely will work on this then.

eliotmiranda:
@johnmci can you review this now?

PR #375, opened 2019-03-06
packaging: Fully catch-up and get it buildable

zecke:
* Add another editpharoinstall.sh that doesn't try to copy from /usr/src
* Match "Name" and spec filename for OpenSUSE tumbleweed
* Fix the PharoV60.sources dependency. It is still 6 and not 7.


estebanlm:
issues should not be related

PR #381, opened 2019-03-21
Fixing the change to fullscreen

tesonep:
Changing the implementation of the fullScreen primitive to follow the guidelines here:

https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/FullScreenApp/FullScreenApp.html

It works from the primitive and from the green button in the window

### DIFF
--- a/platforms/iOS/vm/OSX/sqSqueakOSXOpenGLView.m
+++ b/platforms/iOS/vm/OSX/sqSqueakOSXOpenGLView.m
@@ -1114,27 +1114,20 @@ lastSeenKeyBoardModifierDetails,dragInProgress,dragCount,dragItems,windowLogic,l
 
 - (void)  ioSetFullScreen: (sqInt) fullScreen {
 
-	if ([self isInFullScreenMode] == YES && (fullScreen == 1))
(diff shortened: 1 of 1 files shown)


johnmci (commented):
The apple documents are not clear, one is a test for the state of the view at the moment, the other is a test to see if the window is capable of full screen?  
@property(getter=isInFullScreenMode, readonly) BOOL inFullScreenMode;
The value of this property is YES when the view is in full screen mode or NO when it is not.

 NSFullScreenWindowMask
The window can appear full screen. A fullscreen window does not draw its title bar,
 and may have special handling for its toolbar. 
This mask is automatically toggled when toggleFullScreen: is called.

eliotmiranda (approved):

eliotmiranda:
Appears to work perfectly.  Thanks Pablo!

StephanEggermont:
Should that be fixed in pharo.cog.spur_macos64x64_201903251926?. I still manage to get an unresponsive Pharo image when doing in a playground: Display toggleFullscreen

PR #382, opened 2019-03-25
Initial port of the B3DAccelerator plugin, plus extra Metal and rendering clean up

ronsaldo:
This is the branch with the port of the B3DAccelerator plugin to Metal. Most the examples in Balloon3D are working with this new plugin, after forcing it by modifying:

B3DSceneMorph >> accelerationEnabled
	"Return true if hardware acceleration is enabled"
	"self accelerationSuspended ifTrue:[^false].
	^self valueOfProperty: #accelerationEnabled ifAbsent:[false]"
	^ true

Shaders are now compiled in runtime, from source code which is embedded in a C-string.

In addition to these changes, in the last commit I did some refactoring of the Metal VM code. I am copying the message of that commit:

- I introduced a hack to select the type of rendering view in runtime.
- I am compiling by again the old core graphics based renderer, but this is only used when the -core-graphics option is passed in the command line.
- I added dummy headless view, that does not render and stubs most of the event.
- I added the -metal, -opengl, and -core-graphics command line options for selecting the rendering backend.
- I am testing the Metal implementation by compiling the shader library. If the shader library compilation fails, then Metal will not be used, and instead OpenGL will be used as a fallback, by default.


eliotmiranda:
When I try and compile on 64-bit (build.macos64x64/squeak.cog.spur; macOS High Sierra 10.13.6; clang --version: Apple LLVM version 7.0.0 (clang-700.1.76) Target: x86_64-apple-darwin17.7.0) I get the following error message:

../../platforms/iOS/vm/OSX/sqSqueakOSXMetalView.m:79:17: error: ARC forbids Objective-C objects in struct

that line is the "texture" declaration in

```
typedef struct ExtraLayer
{
    id<MTLTexture> texture;
    int x, y;
    int w, h;
} ExtraLayer;
```

maenu:
Not sure if this fits here, but if I download the latest Pharo 7 VM on macOS High Sierra, the VM seems not to start properly, resulting in a black window and:

```
2019-03-27 11:48:46.282 Pharo[1041:7933912] Pipeline state creation error: Function screenQuadFragmentShader is using language version 2.1 which is incompatible with this OS.
```

I looked into this a while ago, and there should be some flags to enable Metal 1 compatibility, then we would get at least down to El Capitan support. Adding Metal as required dependency is a non-backwards-compatible change, this should be handled with care (e.g. do we need non-metal builds, archive old non-metal builds for older OSs?).

eliotmiranda:
Hi @maenu,

    can you work with @ronsaldo to look at getting down to El Capitan support?  Right now I'm on High Sierra (10.3) and Metal 2 doesn't even work there-on.

maenu:
I will have a look, maybe @ronsaldo is even at PharoDays next week?

maenu:
Sorry to revive the merged thread, but what is the state on this?

I just built the macOS 64bit Pharo Cog VM on Sierra, Metal seems to be enabled (`-DUSE_METAL=1`), and the VM runs my image. But downloading a fresh VM from Pharo downloads gives me `2019-03-27 11:48:46.282 Pharo[1041:7933912] Pipeline state creation error: Function screenQuadFragmentShader is using language version 2.1 which is incompatible with this OS.` again. 

My current hypothesis is, that the downloaded VM is built on a newer macOS, which removes the compatibility for Sierra. Anybody knows how this happens and how we could trick a newer macOS in building a VM with a El Capitan/Sierra/High Sierra compatibility?

The only thing I found to control compatibility is the `-mmacosx-version-min`, which is `10.7` in my local build. It looks like there is nothing other than this flag to achieve compatibility, see also https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/build.rs.

ronsaldo:
@maenu That is interesting. Did you download that VM from bintray? That also looks like an error from the older version.

maenu:
@ronsaldo I got the VM with `curl https://get.pharo.org/64/70+vmLatest | bash`. Maybe that is not really the latest VM.

PR #384, opened 2019-03-27
Fix compilation problem with ARC in some compilers.

ronsaldo:
This fixes a compilation error that is produced by some versions of clang. No ARC pointers in structs are allowed.


eliotmiranda (approved):
Works, but only with the 10.12 SDK (i.e. -metal doesn't crash on startup; still doesn't render correctly).
So for now we're targeting the 10.12 SDK.

PR #390, opened 2019-04-11
fix libgit2 makefile

maenu:
I am using MacPort's CMake which works fine, except that it needs a path argument for the libgit2 makefile. Can somebody check if this is compatible with their CMakes?


krono:
Since this doesnt do harm, as that part is essentially unused, just :shipit:

PR #395, opened 2019-05-04
Add --quiet option to allow an app to own stdout and stderr

zecke:
I envison to have --quiet and --debugoutput FILENAME as options in the future. For an end-user it is difficult to separate application output/error from VM one (e.g. the infamous thread priority warning).

### DIFF
--- a/platforms/unix/vm/sqUnixMain.c
+++ b/platforms/unix/vm/sqUnixMain.c
@@ -643,13 +643,13 @@ static void emergencyDump(int quit)
   printf("\nMost recent primitives\n");
   dumpPrimTraceLog();
 #endif
-  fprintf(stderr, "\n");
+  fprintf(VM_ERR(), "\n");
   printCallStack();
-  fprintf(stderr, "\nTo recover valuable content from this image:\n");
-  fprintf(stderr, "    %s %s\n", exeName, imageName);
-  fprintf(stderr, "and then evaluate\n");
-  fprintf(stderr, "    Smalltalk processStartUpList: true\n");
-  fprintf(stderr, "in a workspace.  DESTROY the dumped image after recovering content!");
+  fprintf(VM_ERR(), "\nTo recover valuable content from this image:\n");
+  fprintf(VM_ERR(), "    %s %s\n", exeName, imageName);
+  fprintf(VM_ERR(), "and then evaluate\n");
+  fprintf(VM_ERR(), "    Smalltalk processStartUpList: true\n");
+  fprintf(VM_ERR(), "in a workspace.  DESTROY the dumped image after recovering content!");
 
   if (quit) abort();
   strncpy(imageName, savedName, sizeof(imageName));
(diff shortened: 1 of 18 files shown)


platforms/unix/vm/sqUnixMain.c
```diff
@@ -863,13 +863,13 @@ sqInt  primitivePluginRequestState(void)	{ return dpy->primitivePluginRequestSta
 
 static void outOfMemory(void)
 {
-  /* pushing stderr outputs the error report on stderr instead of stdout */
+  /* pushing VM_ERR() outputs the error report on VM_ERR() instead of stdout */
```
bencoman, line 866: What is the significance of changing this comment?  (I'm not familiar with *pushOutputFile*)

platforms/unix/vm/sqUnixMain.c
```diff
@@ -863,13 +863,13 @@ sqInt  primitivePluginRequestState(void)	{ return dpy->primitivePluginRequestSta
...
   pushOutputFile((char *)STDERR_FILENO);
   error("out of memory\n");
 }
 
 /* Print an error message, possibly a stack trace, do /not/ exit.
- * Allows e.g. writing to a log file and stderr.
+ * Allows e.g. writing to a log file and VM_ERR().
```
bencoman, line 872: Same again. Search/replaced within comment useful?

bencoman:
An expansive scattering of changes, but pretty much all of them are...
simply refactoring `fprintf(stderr,`  ==>  `fprintf(VM_ERR(),`
excepting the following which looks fine...
```
static FILE *VM_ERR_FILE = NULL;

FILE *VM_ERR(void)
{
	if (!VM_ERR_FILE) {
		VM_ERR_FILE = stderr;
	}
	return VM_ERR_FILE;
}

void sqSetVmErrFile(FILE *file)
{
	VM_ERR_FILE = file;
}



static int vm_parseArgument(int argc, char **argv)
{
   ...
   else if (!strcmp(argv[0], VMOPTION("quiet"))) {
       sqSetVmErrFile(fopen("/dev/null", "w"));
       return 1;
    }
    return 0; /* option not recognised */
}

(void) printUsage {
        ...
	printf("  "VMOPTION("quiet")"                don't print debugging messages from the VM\n");
        ...
}

        
```

And the following seems correct to align with sq.h...
```
- extern void addIdleUsecs(long idleUsecs);
+ extern void addIdleUsecs(sqInt idleUsecs);
```
https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/83aebb122ca2fd32a86ef8b7d478ed32b12dc6b1/platforms/Cross/vm/sq.h#L180

eliotmiranda:
Hi Holger,

    I don't object to this in principle.  But I question whether it is necessary, except on Windows.  Let me ask you why using the shell to redirect stderr to /dev/null is inadequate.  I would simply do

bin/squeak my image.image 2>/dev/null &

Where does this not work?

zecke:
It takes the ability from an application author to use stderr. I look at it from the point of view of using the "ecstatic" cli application (pillar to html) and the scheduler param warning printed by the VM really takes away from the quality/usefulness of the app (I have neither seen this with python, ruby, java, libstdc++ runtimes).

There are probably more elegant ways to achieve this but wanted to get a discussion started.

zecke:
Closing as contributing seemed too hard.

Issue #385, opened 2019-03-27
32-bit libB3DAcceleratorPlugin.dylib fails to link; 32-bit Squeak B3DAcceleratorPlugin links against metal.


eliotmiranda:
Seems to be something strange with the B3DAcceleratorPlugin on 32-bit macos.  The pharo libB3DAcceleratorPlugin.dylib fails to link missing:
  createOpenGLTextureLayerHandle
  destroyOpenGLTextureLayerHandle
  getMainWindowOpenGLContext
  setOpenGLTextureLayerContent

The Squeak one links, but apparently against Metal, which isn't going to work since Metal is 64-bit only (or so I've been told).

nicolas-cellier-aka-nice:
I think that this can be closed because fixed by Ronie.

PR #399, opened 2019-06-08
Fix 32 bits CI builds because of the Metal integration

ronsaldo:
This should fix this issue: https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/385#issue-426205672

On 32 bits it is possible to link with the Metal libraries, but the MTLCreateSystemDefaultDevice() function always returns nil, at least on my machines.


fniephaus:
Thanks for the fix, @ronsaldo!

Issue #396, opened 2019-05-27
Linux Squeak VM X11 keyboard changes breaks 32b VM


kksubbu:
Tim Johnson reported a problem with 32b Linux builds:
 http://forum.world.st/Linux-Squeak-VM-X11-keyboard-changes-td5099584.html

Pressing ctrl-d now acts like debugIt instead of doIt like in the 64b. Ctrl-p also stopping working.

I did a bisection of all versions in https://bintray.com/opensmalltalk/vm/cog/ and narrowed down the problem to

good squeak.cog.spur_linux32x86_201811272342.tar.gz
bad squeak.cog.spur_linux32x86_201812301551.tar.gz

The error is consistent on both 18231 and the oldest 18221 released image for 32b VM but does not appear with 64b VMs.

tcj:
Thanks for all the work you've done on this.  
I (Tim J who reported the problem) have built a debug build and uploaded it to a personal web server.  If you trust a strange binary built on a DigitalOcean server, it is [here](http://www.sonic.net/~digit/a045ab9f1ef59-linux32x86-debug.tgz) (tarball is named after the commit hash).

kksubbu:
Thanks Tim,

The following lines in sqUnixX11.c doesn't seem right. When Ctrl and p are pressed at the same time, sq2KeyPlain should return ctrl-p (16r10). This code returns p instead (16r70). I don't understand why this change was made.
```
2065   if (charCode >= 1 && charCode <= 26) {
2066     /* check for Ctrl-letter that gets translated into charCode 1-26 instead of letters a-z */
2067     KeySym keysym = *symbolic;
2068     if (keysym >= XK_a && keysym <= XK_z)
2069       return (int)'a' + (keysym - XK_a);
2070     if (keysym >= XK_A && keysym <= XK_Z)
2071       return (int)'A' + (keysym - XK_A);
```
Could you try a build with this section removed and see if it fixes your problem?

krono:
Relevant commit: 1b837f94e96b93cb4d117ccdd010825d76dcde57

kksubbu:
Tim,

I used gdb to nop out above code in the text segment as this was quicker and easier than trying to rebuild a 32b binary. Then the image worked like charm.

[...]

PR #400, opened 2019-06-09
Elided code that caused ctrl keyboard shortcuts to fail on 32b images

kksubbu:
This patch solves the problem discussed in https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/396

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -2062,14 +2062,6 @@ static int x2sqKeyPlain(XKeyEvent *xevt, KeySym *symbolic)
       return -1;	/* unknown key */
   if ((charCode == 127) && mapDelBs)
     charCode= 8;
-  if (charCode >= 1 && charCode <= 26) {
-    /* check for Ctrl-letter that gets translated into charCode 1-26 instead of letters a-z */
-    KeySym keysym = *symbolic;
-    if (keysym >= XK_a && keysym <= XK_z)
-      return (int)'a' + (keysym - XK_a);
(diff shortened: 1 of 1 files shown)


krono:
I think we really should wait with this until Eliot's approval. He put the code in there not too long ago, and I'm sure he had good reason to…

nicolas-cellier-aka-nice:
I agree with Tobias, I suggest that we analyze the failure further, specifically what happens to modifiers state before reverting those lines. I commented Issue #396 with a bit more information about my yet incomplete understanding of the problem.

eliotmiranda:
On Sun, Jun 9, 2019 at 8:33 AM Tobias Pape <notifications@github.com> wrote:

> I think we really should wait with this until Eliot's approval. He put the
> code in there not too long ago, and I'm sure he had good reason to…
>

IIRC, I just rubber stamped someone else's contribution.

_,,,^..^,,,_
best, Eliot

akgrant43:
I think this is the relevant issue: https://github.com/feenkcom/gtoolkit/issues/9#issuecomment-447487311
See John Brant's comments from 14 December.

bencoman:
Good find @akgrant43. Paraphrasing feenkcom/gtoolkit#9...

[Edit: ] Better to record this on the issue... https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/396#issuecomment-501308271

kksubbu:
John Brant reported that this patch breaks ctrl-a handling in 64b image. So this patch will not solve the issue for all Squeak and Pharo images. I will withdraw the patch. Further discussions can continue in #396

PR #404, opened 2019-06-13
Fix the minheadless VM for CI

ronsaldo:
This pull request fixes the CMake based build system for the minheadless for building on CI for both, Pharo and Squeak. This also adds two variants of the minheadless vm:

- The true headless VM variant.
- A vm with SDL2 based display that supports standard Pharo and Squeak images, for 32 and 64 bits in Linux, Mac and Windows.

In addition to these fixes, this PR implements adds an image selection dialog for Mac and Windows when the Minheadless VM can not find automatically an image. This allows double clicking the Minheadless VM.

The automatic image search and detection in the Minheadless vm works like this now:
- If there is an image called startup.image in the same folder as the vm executable, (or in the bundle Resources folder in the case of Mac OS X), then this image is executed always.
- If there is not an startup.image file in these folder, the vm executable, the Mac Resource folder, or the working directory is searched for all the files with the .image extension. If only one of these files is found, then this image file is executed. If no images are found, or more than one image is found, then image selection dialog is display on Mac and Windows (on Linux, the VM just exits).

This pull request re-implements the thirdparty dependency download, configuration and build using CMake. Only the dependencies that are actually used by a VM variant are built automatically, and it is built only once. AppVeyor and Travis caches are also supported by the CMake based build system. The thirdparty dependencies that are built with CMake are placed under the .thirdparty-cache/cmake folder to avoid clashing and breaking the Makefile based build system. Building thirdparty dependencies with CMake works, but it is messy and hacky (so it is using the Makefiles). The thirdparty dependency build system should be moved into a separate repository. Third party dependencies are quite stable and do not change as frequent as the VM, so it should be easier, cleaner and faster to handle them in a separate repository.


ronsaldo:
Travis errors are not related to this PR:

curl -o ../../.thirdparty-cache/libgit2-v0.25.1.tar.gz -L https://github.com/libgit2/libgit2/archive/v0.25.1.tar.gz 

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current

                                 Dload  Upload   Total   Spent    Left  Speed


  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (6) Could not resolve host: github.com

make: *** [../../.thirdparty-cache/libgit2-v0.25.1.tar.gz] Error 6

PR #411, opened 2019-06-27
Tag the minheadless builds with CMake with a suffix instead of a prefix.

ronsaldo:
I just noticed that the bintray deployment script has the following regex:

{"includePattern": "../products/((newspeak|pharo|squeak).*\\.(?:gz|zip|dmg))", "uploadPattern": "$1"}

Since I was tagging the minheadless with a cmake-minhdls- prefix, they build artifacts are not matching this regex. I prefer changing the prefix with a -cmake-minhdls suffix instead of messing with a potentially dangerous regex.


fniephaus:
+1 for consistency

PR #412, opened 2019-07-14
Creation script fixes

aletg:
Two little fixes for spur-vmmaker image creation scripts.


nicolas-cellier-aka-nice:
Sounds good, but would merit a [skip ci] or [ci skip] in commit message to avoid starting a new build

PR #419, opened 2019-08-21
Minheadless SqueakSSL plugin compilation bug fix

ronsaldo:
I fixed a mistake in the compilation of the SqueakSSL plugin in the minheadless VM for Mac.

### DIFF
--- a/cmake/PluginsCommon.cmake
+++ b/cmake/PluginsCommon.cmake
@@ -96,6 +96,10 @@ if(DARWIN)
 endif()
 
 # Squeak SSL plugin
+# In the case of the minheadless VM for OS X, we are treating the VM as it were
+# an unix since we are removing all of the platform specific windowing code in
+# this VM variant. For this reason, in the cases where OS X is different than
+# another unix, the plugin code has to be added manually.
 if (APPLE)
 	set(SqueakSSL_Sources
 	    "platforms/iOS/plugins/SqueakSSL/sqMacSSL.c"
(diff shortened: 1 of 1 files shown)


krono:
Please explain.

ronsaldo:
On the minheadless VM that is built using cmake, the add_vm_plugin_sources cmake macro requires specifying the the plugin sources explicitly, which is omitting the src/plugins/SqueakSSL/SqueakSSL.c in the compilation of the plugin. By using the other macro (add_vm_plugin_auto), the platform specific files are automatically found with a glob pattern. This is a mistake that I introduced myself. This problem can be reproduced in Pharo using the minheadless vm of this repository on OS X with the following script:

´´´smalltalk
url := 'https://google.com' asZnUrl.
ZnClient new
    url: url;
    get;
    response
´´´

CI errors on travis are not related to this PR.

krono:
Ok, but then I don't understand why Mac needs the explicit `${SqueakSSL_Sources}`. Apparently, Windows and Unix don't, but also have such specific sources.

ronsaldo:
In the case of the minheadless VM for OS X, we are treating the VM as it were an unix since we are removing all of the platform specific windowing code in this VM variant. For this reason, in the cases where OS X is different than another unix, the plugin code has to be added manually.

ronsaldo:
In the case of the minheadless VM for OS X, we are treating the VM as it were an unix since we are removing all of the platform specific windowing code in this VM variant. For this reason, in the cases where OS X is different than another unix, the plugin code has to be added manually.

fniephaus:
Side note: I would've liked to see the explanation to be part of the commit message. It'll be rather hard to find this discussion if we need to in the future. But it's not too late: we can still add the info to the merge commit.

PR #424, opened 2019-09-03
Restore support for large cursor on linux X11 VM

nicolas-cellier-aka-nice:
Large cursor support is present but optional on X11
It is triggered by compiler define HAVE_LIBXRENDER
It depends on availability of Xrender library which is an optional X extension
Try to test presence of header and library in configure macro
Note that it is also necessary to install the proper package
e.g. on debian flavours

	sudo apt-get install libxrender-dev

### DIFF
--- a/platforms/unix/config/aclocal.m4
+++ b/platforms/unix/config/aclocal.m4
@@ -1,6 +1,6 @@
-# generated automatically by aclocal 1.15 -*- Autoconf -*-
+# generated automatically by aclocal 1.15.1 -*- Autoconf -*-
 
-# Copyright (C) 1996-2014 Free Software Foundation, Inc.
+# Copyright (C) 1996-2017 Free Software Foundation, Inc.
 
 # This file is free software; the Free Software Foundation
 # gives unlimited permission to copy and/or distribute it,
--- a/platforms/unix/config/config.h.in
+++ b/platforms/unix/config/config.h.in
@@ -119,6 +119,9 @@
 /* Define to 1 if you have the <libutil.h> header file. */
 #undef HAVE_LIBUTIL_H
 
+/* Have Xrender library */
+#undef HAVE_LIBXRENDER
+
 /* Define to 1 if you have the <linux/fb.h> header file. */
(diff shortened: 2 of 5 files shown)


nicolas-cellier-aka-nice:
Could someone fluent in autotools double check this
It worked for me, but I'm very unsure.

Should I commit the generated aclocal.m4 config.h.in and configure files?
I presume so...

nicolas-cellier-aka-nice:
I committed the generated configure. Of course it works for me, the question is if it works for everyone.

I have a doubt, it seems that I set the `-DHAVE_LIBXRENDER` define if I find the header,
then check whether the library libXrender.so is required or not (and add it to `LIBS` if required).

That does not cover the case when header would be present, but the library would be required AND absent. Should we cover that case?

nicolas-cellier-aka-nice:
Ok, not good. That means that the VM shall attempt to link it dynamically with dlopen, and issue a message on console if that fails... on ubuntu, libXrender is an optional package, i guess similar in other distros... 

Or we could document the requirement and add the package dependency in squeak/pharo/cuis/newspeak packages...

Or we could attempt to link statically, which would be somehow simple and robust, but not secure (no update).

Or we could find another solution for displaying larger cursors. How do others apps handle those highres Screens?

dtlewis290:
The only real problem is that the VM is not issuing an understandable error message in the case of a missing runtime library. For example, if the VM module-display-X11 module requires runtime library libXrender, and if libXrender.so is not installed, then a message like this would be sufficient:

$ squeak myImage
failed loading module /usr/local/lib/squeak/4.16.6-3772/so.vm-display-X11 due to libXrender.so.1: cannot open shared object file: No such file or directory
could not find display driver vm-display-X11; either:
  - check that /usr/local/lib/squeak/4.16.6-3772/vm-display-X11.so exists, or
  - use the '-plugins <path>' option to tell me where it is, or
  - remove DISPLAY from your environment.
Aborted (core dumped)

I can make this work on interpreter VM (SVN trunk) but module loading seems to have changed on oscog and I have not yet been able to make it work here. More to follow.

PR #431, opened 2019-10-16
Latest linux vm cant handle vm display null

### LINKED ISSUE #421: Latest Linux VM can't handle -vm-display-null
On Linux (Debian 10, Ubuntu 14.04.5) `curl get.pharo.org/64/vmLatest80 | bash; ./pharo` fails with
> could not find module vm-display-null

theseion:
Fixes #421

### DIFF
--- a/scripts/ci/travis_install.sh
+++ b/scripts/ci/travis_install.sh
@@ -16,7 +16,12 @@ if [[ "${ARCH}" = "linux64x64" ]]; then
             libpulse-dev \
             libaudio-dev \
             gcc-multilib \
-            uuid-dev
+            uuid-dev \
+            libglu1-mesa-dev \
+            autoconf \
+            libtool \
+            curl \
+            cmake
 elif [[ "${ARCH}" = "linux32x86" ]]; then
     sudo apt-get update -y
     sudo apt-get remove -q -y gvfs-daemons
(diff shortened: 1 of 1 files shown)


theseion:
I'm aware that cmake is probably only used by Pharo builds but the overhead to install it is small and there's no facility yet to have different dependencies for different flavors.

fniephaus:
Thanks!

theseion:
Great, thanks! Now SmalltalkCI should start working again for Pharo 8.

fniephaus:
> Great, thanks! Now SmalltalkCI should start working again for Pharo 8.

For that, @estebanlm and the Pharo VM team needs to push a new version of the VM to files.pharo.org, right?

theseion:
Yes. There's also a separate issue: https://github.com/pharo-project/opensmalltalk-vm/issues/15

estebanlm:
thing is, we cannot push a new version to stable because the branch is broken for MacOS (due the changes to make it work in metal, I think). The VM is not compatible with older versions and then we are kind of stuck there :/

fniephaus:
@estebanlm so what's the plan on your side? Revert the Metal backend for now? Providing two flavors (e.g. `macOS-new` and `macOS-legacy`)?

theseion:
Just to clear up a possible confusion: the build updates the *latest* VM on files.pharo.org automatically, so Pharo 8 alpha builds with SmalltalkCI work (apart from the new problem with the segfault: https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/433).

PR #432, opened 2019-10-16
x11: emulate only up/down scroll events via keyboard, not left/right

tom95:
This restores the behavior to what we previously had and fixes an
annoying side effect, where vertical scrolling on a touchpad causing
occassional horizontal scrolling events would move the cursor in a text
editor, since we generate ctrl+left/right.


fniephaus:
Thanks!

PR #438, opened 2019-10-24
Compile newspeak with clang

nicolas-cellier-aka-nice:
This restore newspeak sanity on linux64x64 and Ci sanity by authorizing squeak.cog.v3 failures


nicolas-cellier-aka-nice:
One new type of failure for each and every day, that's remarkable!
This time 5 tests did timeout on squeak.cog.spur macos32x86...

nicolas-cellier-aka-nice:
Hi Tim,
no, it would then mean that we have faster feedback and would steer the software less blindly.
Like it is now, you see some oscillations appearing, allow failure of this brand, retract, redo...
We are at the limit of feedback loop unstability, so near a kind of optimality ;)

PR #446, opened 2019-11-17
fix calculation of x11 scroll distances

tom95:
I believe these values should be like the below.

Refer to for example here: http://xahlee.info/linux/linux_x11_mouse_button_number.html (the first scroll event is button 4). And the up-direction should be negative for Squeak. After these adjustments the behavior was correct on my machine.

To try out, ensure that you have `Smalltalk sendMouseWheelEvents: true`


nicolas-cellier-aka-nice:
This is absolutely right, good find!
I think that we under-use this feature!

PR #458, opened 2019-12-24
Improved local address support on linux

nicolas-cellier-aka-nice:
Do not rely on hardcoded interface names (eth0, wlan0) in sqResolverLocalAddress(void). Use ifa_flags to filter the list instead.
When an interface has no address assigned, just skip it instead of failing the primitive and leaking memory (return without calling freeifaddrs).

On behalf of Levente


nicolas-cellier-aka-nice:
The changes look good to me, I let a bit of time if someone else want to review.

krono:
Just a question, is that Linux-specific flags or do they work, eg, on bsd'en?

(cc'ing @ckeen, asking for a tiny test please :) )

krono:
hooray!

PR #478, opened 2020-02-28
Use NSWindow API for changing fullscreen mode

fniephaus:
instead of going through NSView's API. The latter does offer more options, but messes up the fullscreen mode for some reason ([UI becomes unresponsive because input events are no longer received](http://forum.world.st/ANN-Squeak-5-3-RC1-tp5112435p5112454.html)).

Using `self.window toggleFullScreen:self` also does not mess with additional displays (only the display showing Squeak is put into fullscreen mode).

Please test and review!

### DIFF
--- a/platforms/iOS/vm/OSX/sqSqueakOSXMetalView.m
+++ b/platforms/iOS/vm/OSX/sqSqueakOSXMetalView.m
@@ -48,7 +48,7 @@
 
 extern SqueakOSXAppDelegate *gDelegateApp;
 extern struct VirtualMachine* interpreterProxy;
-extern int getFullScreenFlag();
+extern sqInt getFullScreenFlag();
 
 static sqSqueakOSXMetalView *mainMetalView;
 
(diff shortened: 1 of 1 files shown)


platforms/iOS/vm/OSX/sqSqueakOSXMetalView.m
```diff
@@ -48,6 +48,7 @@
 
 extern SqueakOSXAppDelegate *gDelegateApp;
 extern struct VirtualMachine* interpreterProxy;
+extern int getFullScreenFlag();
```
nicolas-cellier-aka-nice: Just a detail, but should be sqInt

fniephaus: Have you given the PR a try locally?

eliotmiranda (commented):
Personally I loathe and despise (fullScreen == 0)  Correct C is !fullScreen.  ;-)

And getFullScreenFlag() == (fullScreen == 1) is better written as getFullScreenFlag() == (fullScreen != 0) (or occasionally one will see getFullScreenFlag() == !!fullScreen).

eliotmiranda (approved):

PR #481, opened 2020-04-07
Improve unix portability

krono:
today: FreeBSD


krono:
The interesting builds pass. :shipit:

Issue #455, opened 2019-12-14
OpenSSL 1.0.2 end of support from 1 Jan 2020


PierceNg:
As per https://www.openssl.org/policies/releasestrat.html.

My fork has an [update3p branch](https://github.com/PierceNg/opensmalltalk-vm/tree/pierce_update3p) that builds with OpenSSL 1.1.1d and also related minor fixes for one or some of the Pharo mvm.

eliotmiranda:
Hi Pierce,

On Fri, Dec 13, 2019 at 5:29 PM Pierce Ng <notifications@github.com> wrote:

> As per https://www.openssl.org/policies/releasestrat.html.
>
> My fork has an update3p branch
> <https://github.com/PierceNg/opensmalltalk-vm/tree/pierce_update3p> that
> builds with OpenSSL 1.1.1d and also related minor fixes for one or some of
> the Pharo mvm.
>

Thanks!!

Can you coordinate with Tobias to fold this back into the OpenSmallktalk
trunk?  i.e. if Tobias, who is the keeper of OpenSSL says that the code is
good, can you both ensure it gets pushed into OpenSmalltalk, and if there
are issues can you both report back?

_,,,^..^,,,_
best, Eliot

PierceNg:
Will do. What's Tobias GH account name?

krono:
It me.

PierceNg:
@krono For cmake-based Pharo builds, update ```third-party/openssl.spec``` URL and version numbers as below. I don't build the other VMs so I don't know how they specify building OpenSSL.

```
openssl_spec_download_url:=https://www.openssl.org/source/openssl-1.1.1d.tar.gz
openssl_spec_archive_name:=openssl-1.1.1d.tar.gz
openssl_spec_unpack_dir_name:=openssl-1.1.1d
openssl_spec_product1_name_macOS:=libssl.1.1.dylib
openssl_spec_product2_name_macOS:=libcrypto.1.1.dylib
openssl_spec_product1_name_linux:=libssl.so.1.1
openssl_spec_product2_name_linux:=libcrypto.so.1.1
openssl_spec_product1_name_windows:=ssleay32.dll
openssl_spec_product2_name_windows:=libeay32.dll
openssl_spec_symlinks_macOS:=libssl*.dylib libcrypto*.dylib
openssl_spec_symlinks_linux:=libssl.so* libcrypto.so*
```

krono:
[...]

PR #474, opened 2020-02-02
Update OpenSSL download to v1.1.1d as v1.0.2q EOLed 1 Jan 2020.

PierceNg:
Fixes #455

PR #482, opened 2020-04-10
Update OpenSSL download to v1.1.1f as v1.0.2q EOLed 1 Jan 2020

VincentBlondeau:
URL is now https://www.openssl.org/source/openssl-1.1.1f.tar.gz
name of the dlls on windows have changed.
targets for building are updated according to https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/INSTALL
Windows build locally successful but I cannot test the other platforms


VincentBlondeau:
Update of #474 to resolve #455

krono:
VM mostly not affected.

PR #485, opened 2020-04-11
Fix #484 [Win64]Cannot save and load image files with a heap whose size is more than 0xff ff ff ff (~4.1GB)

### LINKED ISSUE #484: [Win64]Cannot save and load image files with a heap whose size is more than 0xff ff ff ff (~4.1GB)
The windows native functions used to deal with the files (ReadFile WriteFile) are limited to write and read bytes defined by a length argument. The length argument is a DWORD and then limited to 0xff ff ff ff .

Images with a size of more than ~4.1GB are not saved properly and cannot be open again.

Fix incoming soon!

VincentBlondeau:
The problem is actually when there is only one segment of memory of more of 0xff ff ff ff to read or to write to disk.

This fix the issue and it is tested with a real case on windows.

I added some checks on function return.

Let me know if changes are to be done to follows some conventions or optimizations.

### DIFF
--- a/platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
+++ b/platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
@@ -512,32 +512,26 @@ squeakFileOffsetType sqImageFilePosition(sqImageFile h)
 size_t sqImageFileRead(void *ptr, size_t sz, size_t count, sqImageFile h)
 {
   DWORD dwReallyRead;
-
   size_t reallyRead = 0;
-
   size_t totalToRead = count * sz;
-
   squeakFileOffsetType position;
 
-  if (totalToRead > (size_t)MAX_DWORD - 1) {
-    while (reallyRead != totalToRead) {
-      DWORD toRead = (totalToRead - reallyRead) > (size_t)MAX_DWORD ? MAX_DWORD : totalToRead - reallyRead;
-      ReadFile((HANDLE)(h - 1), (LPVOID)((sqInt)ptr + (sqInt)reallyRead), toRead, &dwReallyRead, NULL);
-      reallyRead += dwReallyRead;
-    }
-    return (reallyRead / sz);
-  }
-
   position = sqImageFilePosition(h);
-  ReadFile((HANDLE)(h - 1), (LPVOID)ptr, count * sz, &dwReallyRead, NULL);
-  while(dwReallyRead != (DWORD)(count*sz)) {
-    DWORD err = GetLastError();
-    if(sqMessageBox(MB_ABORTRETRYIGNORE, TEXT("Squeak Warning"),TEXT("Image file read problem (%d out of %d bytes read)"), dwReallyRead, count*sz)
-       == IDABORT) return (dwReallyRead / sz);
-    sqImageFileSeek(h, position);
-    ReadFile((HANDLE)(h-1), (LPVOID) ptr, count*sz, &dwReallyRead, NULL);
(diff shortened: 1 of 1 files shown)


platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
```diff
@@ -37,6 +37,7 @@ extern struct VirtualMachine *interpreterProxy;
 
 #define true  1
 #define false 0
+static const DWORD MAX_DWORD = 4294967295;
```
j4yk, line 40:
```suggestion
static const DWORD MAX_DWORD = MAXDWORD;
```

j4yk: ...so really one could just use the constant provided by the Windows headers below.

platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
```diff
@@ -511,18 +512,26 @@ squeakFileOffsetType sqImageFilePosition(sqImageFile h)
...
-    if(sqMessageBox(MB_ABORTRETRYIGNORE, TEXT("Squeak Warning"),TEXT("Image file read problem (%d out of %d bytes read)"), dwReallyRead, count*sz)
-       == IDABORT) return (dwReallyRead / sz);
-    sqImageFileSeek(h, position);
-    ReadFile((HANDLE)(h-1), (LPVOID) ptr, count*sz, &dwReallyRead, NULL);
+  while (reallyRead != totalToRead) {
+    DWORD toRead = (totalToRead - reallyRead) > (size_t)MAX_DWORD ? MAX_DWORD : totalToRead - reallyRead;
+    BOOL ret = ReadFile((HANDLE)(h - 1), (LPVOID)((sqInt)ptr + (sqInt)reallyRead), toRead, &dwReallyRead, NULL);
+    reallyRead += dwReallyRead;
+
+    if (!ret | dwReallyRead != toRead) {
```
j4yk, line 525: Not sure whether this is a bug or "clever" code to use the bitwise or. I suggest using the boolean operators (&& and ||) instead for clarity.

platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
```diff
@@ -543,9 +552,26 @@ squeakFileOffsetType sqImageFileSeekEnd(sqImageFile h, squeakFileOffsetType pos)
...
+  size_t totalToWrite = count * sz;
+  squeakFileOffsetType position;
+
+  position = sqImageFilePosition(h);
+  while (reallyWritten != totalToWrite) {
+    DWORD toWrite = (totalToWrite - reallyWritten) > (size_t) MAX_DWORD ? MAX_DWORD : totalToWrite - reallyWritten;
+    BOOL ret = WriteFile((HANDLE)(h - 1), (LPVOID)((sqInt)ptr + (sqInt) reallyWritten), toWrite, &dwReallyWritten, NULL);
+    reallyWritten += dwReallyWritten;
+
+    if (!ret | dwReallyWritten != toWrite) {
```
j4yk, line 566: See above.

eliotmiranda (approved):

platforms/win32/plugins/FilePlugin/sqWin32FilePrims.c
VincentBlondeau: It is not clever code here. Thanks for spotting it!

PR #490, opened 2020-04-22
SunOS

cstes:
Solaris fixes


krono:
Here's a crazy idea: can we fold this into the linux build dirs and have the sunos dir be a symlink? 

The specifica (like openssl no dynamic) could be dealt with in configure and the platform itself :)

kksubbu:
Good idea! I would even go further and drop sunos directory altogether. Builds should be scripted based on technology rather than brands. If the variant can no longer be handled by parsing environment, then sunos can be carved into a separate directory.

eliotmiranda (approved):
Please add the build directories when convenient.  I undid inadvertent changes to build.linux64x64/squeak.cog.spur/build.itimerheartbeat/mvm.  I hope you don' mind :-)

PR #494, opened 2020-04-30
new Sun pull request

cstes:
Hash: SHA256

I've made some changes to support Solaris / SunOS.

First of all, I've removed my modification of the "default" pulseaudio choice in the case of #if sun;
the reason is that there was concern about this on other platforms.

And additionally it is possible to setup the Solaris build environment so that the header files soundcard.h sys/soundcard.h and  sys/audioio.h are not found (are absent).

In that case the configure script will think it does not have build OSS audio and Sun modules, and the only module that will be built is "vm-sound-pulse".   So the system will load the only module available : pulseaudio.

So no change is really required to get the result that "pulseaudio" is now the default on Solaris.

Please have a look at the proposed changes on the "sun" branch and hopefully this can be integrated.

Thanks,
David Stes

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJeqojuAAoJEAwpOKXMq1MaM9AH/0FPPQdg9EpvoV7LWhAWAs+i
7CxEWT6vhoRFuAXB73Op3rtQEmLW+sQtS4v3V4+F9KzFhhHpj0xlI9ZL7Ua52WH/
M9YwoFyct1yZxP00hbB9oSxnEw7G2JrG8CmCYjPKcNUSUUbeRWtoV34POcGWrIuJ
0+NgBD2oDyhmRDZnR+5d5MnZWGFkQCoxCeWu2PcN1t+fO+CWhbaLdE7DUgY4nzdR
ETkR323jFrqqzC9yXvn7WPab9E9lOsjQkzx4+2U9p4zrVphtPkYsX1SX7ngunCny
rn7Qw8yWJsxMtwagHDD0ttZXwvXWdts/iSBnKiRDPr/gU1qfUtlKujTjILr+39M=
=7/ro
-----END PGP SIGNATURE-----


eliotmiranda (approved):
Looks great.  Later on we could try and define min & max more centrally.  I like include_ucontext.h a lot.  Thank you to those involved.

PR #495, opened 2020-04-30
Dump registers on OpenBSD/amd64 platforms

ckeen:
Just a small debugging aid to dump the registers on this platform accordingly


krono:
:shipit:

PR #496, opened 2020-05-07
Sunupdate

cstes:
Hash: SHA256

Hello,

There's two updates for the SunOS port here.

- - Change #define sk_GENERAL_NAME_freefunc for SunPRO C compiler
- - SunOS: add a note in HowToBuild on pkg-config

The first change is for the OpenSSL SqueakSSL case of openssl 1.0.2,
to fix a type cast ... I think this potentially impacts all openssl 1.0.2 platforms that use --disable-dynamicopenssl but it seems OK (I think/hope).

For the second change this is a "documentation" update to the HowToBuild
"HOWTO" documentation in the build.sunos directory; this has no further impact.

Thanks,
David Stes

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJes8ijAAoJEAwpOKXMq1MamV8H/2MrggX6VzRxGCdP40MN5vyE
HuynHVCC68dLneHoYSvaMi4IrVGXS6lYMGa8FfNZsWuVCZDW/rii1nZNH9qo6zCM
6QRekLwKOpp63tXGpKj3C/e8skju6cPvCd9LGWguSqPUVHwFCqOSawYI4OEckvvr
6Nvnqk7ER2yE/iuJSeWgIRLVq4kE/Aq3x8wAtf7zgITHhxr8fEG5+H/Y7icKHW1J
aTv2Z7jDHTcJqM/sEC0la2e8K2iDHkhKBim0K/PS1LtLgzZGAfycdmHJoq0KccRT
gXpEHjsDIEdfvWmIU9xjIiwsIdg5Cz7l5Y7ipb8s1l5cZKv4ct5JnwBl2+Rw+MY=
=oc8N
-----END PGP SIGNATURE-----


cstes:
> Why?
> I don't think that the freefunc is declared as such in the old openssl version.

with the definition 
-#define sk_GENERAL_NAME_freefunc void(*)(void*)

the compile aborts with the SunPro C compiler.

 platforms/unix/plugins/SqueakSSL/sqUnixOpenSSL.inc", line 122: operands have incompatible types:
         pointer to function(pointer to void) returning void ":" pointer to function(pointer to struct GENERAL_NAME_st {int type, union  {..} d}) returning void

With the definition
+#define sk_GENERAL_NAME_freefunc void(*)(GENERAL_NAME*)

compilation works.

The code

  sqo_sk_GENERAL_NAME_pop_free(sANs, (sk_GENERAL_NAME_freefunc)sqo_sk_free);

expanded - with the old defintion of sk_GENERAL_NAME_freefunc -  to:

  sk_pop_free ( ( ( _STACK * ) ( 1 ? ( sANs ) : ( struct stack_st_GENERAL_NAME * ) 0 ) ) , ( ( void ( * ) ( void * ) ) ( ( 1 ? ( ( void ( * ) ( void * ) ) sk_free ) : ( void ( * ) ( GENERAL_NAME * ) ) 0 ) ) ) );

with the new definition of sk_GENERAL_NAME_freefunc it expands to :

                 sk_pop_free ( ( ( _STACK * ) ( 1 ? ( sANs ) : ( struct stack_st_GENERAL_NAME * ) 0 ) ) , ( ( void ( * ) ( void * ) ) ( ( 1 ? ( ( void ( * ) ( GENERAL_NAME * ) ) sk_free ) : ( void ( * ) ( GENERAL_NAME * ) ) 0 ) ) ) );

which the SUNpro compiler accepts.

cstes:
> @cstes Can you have a look at the defined `OPENSSL_VERSION_NUMBER`? The definition to `void*` is necessary in earlier versions of OpenSSL, only later versions have the stacks typed.

The header opensslv.h  has a definition # define OPENSSL_VERSION_NUMBER  0x100020ffL
That header is included by crypto.h.

In my case I observed that I had to change

--- a/platforms/unix/plugins/SqueakSSL/openssl_overlay.h
+++ b/platforms/unix/plugins/SqueakSSL/openssl_overlay.h
@@ -164,7 +164,7 @@
 #define sqo_SSL_library_init SSL_library_init
 #define sqo_SSL_load_error_strings SSL_load_error_strings

-#define sk_GENERAL_NAME_freefunc void(*)(void*)
+#define sk_GENERAL_NAME_freefunc void(*)(GENERAL_NAME*)

krono:
Then it is strange this did not come up earlier 😕

cstes:
> Then it is strange this did not come up earlier 😕

I don't have the issue with "gcc" or with "clang".   Only with the SunPro C compiler.   Technically speaking I find it interesting to compile with the SunPro C compiler although that I can use gcc as an alternative.   By the way, the "alternative" that I also have is to patch the source before compilation, which is also fine by me, then I patch it for the SunPro C specific case.     However I submit the issue "upstream" so to speak to see what you think of it, maybe another solution exists.   I think the change from point to void to point to GENERAL_NAME looked fine to me, by the way that fix also works for me in the gcc and clang cases.

krono:
yeah, you might be right… 
I'm only afraid it might break openssl 0.9.9, but nobody should use that anyway¬
lets keep it

cstes:
Hash: SHA256


This is a reminder about 
https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/496

Currently I apply manually this small patch before building the
OpenSmalltalk VM on Solaris 11.

My understanding from the feedback so far "let's keep it",
this means that it is (hopefully) acceptable to merge it into Cog.

That would be useful as I don't have to manually apply the patch any longer.

It's not a disaster if it cannot be merged into Cog for some reason,
as I can continue to just apply the patch of course,
but I'm submitting the patch with the hope that it can be integrated !


Thanks,
David Stes

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJfIvoPAAoJEAwpOKXMq1Ma8KMIAIVrViG/cNkzqHRp7B0HVCvI
BROJsxFyT5tvfMpB1wI3kj+QmMZ7aK7vnEg2q8kZ5f2rMFV7Zm+xk3gfO1qER1NK
HZV40QBNAld/QDekwK0KJ37hNPN7DBvfp9B4VVdlmX9B6YNbQjOL3TyS1T6jaXbe
XwaiktsQ0OGpWuxrYmjxLOHOkyfoFqJn/CmU6mbU9/wdw4Dh5qX2UKBmoaKXs8KU
FSyx5OS2b5PrXbXqsz0vrkBfevY92kiPY8p5BaGhh+qRvC2fyf/9qmvmHeTFw7aA
TcnKStO+7Cgn9hcE/pNhxrIGuJYLZbOShZkHoUfCsn6d34RPs89dCVgHxD4JqJw=
=+OzJ
-----END PGP SIGNATURE-----

cstes:
Reminder about this patch submitted a few months ago :

Change void * to GENERAL_NAME* in :

-#define sk_GENERAL_NAME_freefunc void()(void)
+#define sk_GENERAL_NAME_freefunc void()(GENERAL_NAME)

krono:
thanks. sorry…

PR #499, opened 2020-05-12
Bump minimal supported Windows version to Windows Vista (0x0600).

marceltaeumel:
Also fixes a minor issue in MSVC Makefile for 64x64, which tried 0x0801 for Windows 8 and 0x1001 for Windows 10. The latter would actually be 0x0A00. See https://docs.microsoft.com/de-de/cpp/porting/modifying-winver-and-win32-winnt

Also in the Makefile for 64x64, define both _WIN32 and _WIN64. See https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros

Also in the Makefile for 64x64, define both WIN32 and WIN64 because those can be used to identify the Windows platform in application code such as in processors/IA32/bochs. I am not aware of any #ifdef WIN64 at the moment. Most code uses #ifdef _WIN64 to check for that.

Note that, having defined both _WIN32 and _WIN64, code for 32-bit and 64-bit must always begin with #ifdef _WIN64 and only then #elif _WIN32.

I think that the MSVC compiler defines _WIN32 and _WIN64 automatically. Yet, now it is documented in the Makefile.

### DIFF
--- a/build.win32x86/common/Makefile.msvc.tools
+++ b/build.win32x86/common/Makefile.msvc.tools
@@ -53,8 +53,8 @@ ifdef MSVC
 MEMACCESS:=-DUSE_INLINE_MEMORY_ACCESSORS=1
 endif
 
-# Set minimum version to WindowsXP (see /cygwin/usr/include/w32api/w32api.h)
-WINVER:=-D_WIN32_WINNT=0x0501 -DWINVER=0x0501
+# Set minimum version to Windows 8 (see /cygwin/usr/include/w32api/w32api.h)
+WINVER:=-D_WIN32_WINNT=0x0602 -DWINVER=0x0602
 
 # define _MT to eliminate the inline versions of printf et al in mingw/stdio.h
 #NOBUILTIN:= -fno-builtin-printf -fno-builtin-putchar -fno-builtin-fprintf
@@ -62,7 +62,8 @@ NOBUILTIN:=-D_MT
 MACHINE:=x86
 
 DEFS:=	$(COGDEFS) $(MEMACCESS) $(WINVER) \
-	-DWIN32=1 -DUNICODE=1 -DWIN32_FILE_SUPPORT -DNO_ISNAN -DNO_SERVICE \
+	-DWIN32=1 -D_WIN32=1 -DUNICODE=1 \
+	-DWIN32_FILE_SUPPORT -DNO_ISNAN -DNO_SERVICE \
 	$(NDEBUG) -DLSB_FIRST -D'VM_NAME="$(VM_NAME)"' $(XDEFS) $(CROQUET)
 
 #############################################################################
--- a/platforms/win32/vm/sqWin32Main.c
+++ b/platforms/win32/vm/sqWin32Main.c
@@ -38,54 +38,10 @@
 #  endif
 #endif
 
-
-/************************************************************************************************************/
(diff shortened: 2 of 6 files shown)


marceltaeumel:
That error on TravisCI is misleading. All builds pass.

build.win32x86/common/Makefile.msvc.tools
```diff
@@ -53,16 +53,17 @@ ifdef MSVC
 MEMACCESS:=-DUSE_INLINE_MEMORY_ACCESSORS=1
 endif
 
-# Set minimum version to WindowsXP (see /cygwin/usr/include/w32api/w32api.h)
-WINVER:=-D_WIN32_WINNT=0x0501 -DWINVER=0x0501
+# Set minimum version to Windows 8 (see /cygwin/usr/include/w32api/w32api.h)
+WINVER:=-D_WIN32_WINNT=0x0602 -DWINVER=0x0602
```
nicolas-cellier-aka-nice, line 57: As suggested on the mailing list, I would just use 0x0600

See also https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2019 for the list of _WIN32_WINNT version.

platforms/win32/vm/sqWin32Main.c
```diff
@@ -925,41 +879,20 @@ sqInt  fileHandleType(HANDLE fdHandle) {
...
-			GetProcAddress(GetModuleHandle(TEXT("kernel32.dll")), "GetFileInformationByHandleEx");
-		if (!pGetFileInformationByHandleEx)
-			return -1;
-	}
-
-	nameinfo = malloc(size);
-	if (nameinfo == NULL) {
-		return -1;
-	}
+	nameinfo = alloca(size);
```
nicolas-cellier-aka-nice: Why not, alloca is a bit shorter, but the pairs of malloc and free seems correct here...
I'd rather consider this change as un-necessary.

krono:
wanna merge? care to first update `build.win64x64/common/Makefile.msvc.tools` ?

build.win32x86/common/Makefile.msvc.tools
eliotmiranda: How would you feel about setting minimum supported version to 8.0?  I want to do this because of the GetSystemTimePreciseAsFileTime function, which appears in 8.0 nd neatly solves the time issues in the win32 VM (use of the millisecond clock to derive greater precision than the 15ms tick in GetSystemTimeAsFileTime causes significant direct over the course of several days).

krono: We can have `0x6000` (Vista) _and_ use that function. Ping me in a week or so, when I'm back from vacation

krono:
We're still targeting XP as of now for 332bit: https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/building/win32x86/common/Makefile.msvc.tools#L59
but already Win8 for 64bit: https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/building/win64x64/common/Makefile.msvc.tools#L52

Anyway, this PR is superseded by reality i think

marceltaeumel:
CI builds use MSYS2 and are thus targeting Windows XP. Only Cygwin+MVSC+Clang is targeting Windows 8 ... by "accident" I suppose as this was never discussed.

PR #513, opened 2020-07-14
Add missing Cryptography plugins to squeak builds

smalltalking:
DESPlugin, MD5Plugin and SHA2Plugin as external plugins

### DIFF
--- a/build.linux32ARMv6/squeak.cog.spur/plugins.ext
+++ b/build.linux32ARMv6/squeak.cog.spur/plugins.ext
@@ -13,9 +13,9 @@ UUIDPlugin \
 WeDoPlugin \
 ImmX11Plugin \
 XDisplayControlPlugin \
-#BochsIA32Plugin \
-#BochsX64Plugin \
-#GdbARMPlugin
 DESPlugin \
 MD5Plugin \
 SHA2Plugin \
+#BochsIA32Plugin \
+#BochsX64Plugin \
+#GdbARMPlugin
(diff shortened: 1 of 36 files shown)


build.linux32ARMv6/squeak.cog.spur/plugins.ext
```diff
@@ -16,3 +16,6 @@ XDisplayControlPlugin \
 #BochsIA32Plugin \
 #BochsX64Plugin \
 #GdbARMPlugin
+DESPlugin \
+MD5Plugin \
+SHA2Plugin \
```
fniephaus, line 18: Not sure these files are allowed to end with an ` \`. More importantly, these three new items must be listed *before* the first `#`. Currently, they are unfortunately all commented out. Yes, the syntax of this file is terrible!

fniephaus (changes requested):

build.linux32ARMv6/squeak.cog.spur/plugins.ext
smalltalking: I wrote a script to rewrite this files en massse because there are so many of them. The script intentionally adds a backslash to the end of all non-comment lines because on linux it just works, and having the \ there has the benefit of easier addition of more lines in the future.
The interesting thing here is that only the 64-bit mac version had problem with the backslash. Neither linux, windows or the 32-bit mac had errors.

On linux, the commented out lines did not cause any problems either. I don't know which program processes these files (and raised the error), but the 64-bit mac version is clearly different that the ones on linux and windows which both seem to accept comments in between lines and the last line ending with \\.

smalltalking: Moving the comments to the end turned out to be enough to fix the errors on 64-bit mac.

PR #515, opened 2020-08-23
vm-display-fbdev

KenDickey:
Note: mvm needs TLC for framebuffer builds.
Armbian does not use MUSL, but -DMUSL was harmless.
ToDo: mouse wheel events not yet converted to arrow codes.
Armbian-Notes.txt and AlpineLinux-Notes.txt need to be migrated or merged into HowToBuild


KenDickey:
HowToBuild should be just a duplicate.  Shoule be ignorable..

eliotmiranda (approved):

PR #523, opened 2020-09-23
configure script : check for <ifaddrs.h>

dcstes:
Hash: SHA256

A few commits that could be of use:

  1. modify configure.ac to check for the header file <ifaddrs.h>
  2. update README of UnicodePlugin
  3. #if COGVM in the sqCogStackAlignment.h for getReturnAddress

The COGVM compiles and works in 32bit on Solaris 10.
It also works now (thanks to code for getReturnAddress) in 64bit on Solaris 11.

Solaris 10 has no header file <ifaddrs.h> it appears, to in that case,
the configure script can automatically 'activate' the old code in
sqUnixSocket.c.  In the Solaris 11 case, it has <ifaddrs.h>, so activate the
new code on Solaris 11.

David Stes

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJfazvaAAoJEAwpOKXMq1MaOBMH/0oAanU940wnZi6UnptEXDzL
H4TGeGlQvX+2939+PP79CxAa0yMNQWe/Ew1f2GsERQruKXZAr76lYQPXaJsY2duI
dUeNaNxvPUuKDlsKnrltUnfDgDazYER8NluCZE3tBV8ZUVxak07u/HLUMs/lXP/9
MO/UctlbDSc8KgZlUkBlaffCzz/DtNV/CrgHZNUlPHb9GyLJD2mctc0H7qYkr1o6
2Y3QoOrC1U//4nK/2JbUV9i4lVVj/g1P9CSjRMe2P1ptZp6GfoWod1XAQMM1TIFP
tjZqhxFDB3PHnhdhyBbLrqEbAXoexi2F6nET3xGXEUPYJr6QL6w9rXf3wdhJGVM=
=tynH
-----END PGP SIGNATURE-----

### DIFF
--- a/platforms/Cross/vm/sqCogStackAlignment.h
+++ b/platforms/Cross/vm/sqCogStackAlignment.h
@@ -17,7 +17,7 @@
 /* getReturnAddress optionally defined here rather than in sqPlatformSpecific.h
  * to reduce duplication. The GCC intrinics are provided by other compilers too.
  */
-#if !defined(getReturnAddress)
+#if COGVM && !defined(getReturnAddress)
 # if _MSC_VER
 #	define getReturnAddress() _ReturnAddress()
 #	include <intrin.h>
--- a/platforms/unix/config/config.h.in
+++ b/platforms/unix/config/config.h.in
@@ -114,6 +114,9 @@
 /* Define to 1 if you have the <iconv.h> header file. */
 #undef HAVE_ICONV_H
 
+/* Define to 1 if you have the <ifaddrs.h> header file. */
+#undef HAVE_IFADDRS_H
+
 /* Interpreter header file present */
 #undef HAVE_INTERP_H
 
--- a/platforms/unix/config/configure.ac
+++ b/platforms/unix/config/configure.ac
@@ -292,6 +292,7 @@ AC_HEADER_STDC
 AC_CHECK_HEADERS_ONCE([unistd.h string.h fcntl.h sys/file.h sys/param.h])
 AC_CHECK_HEADERS_ONCE([sys/time.h sys/filio.h sys/select.h])
 AC_CHECK_HEADERS_ONCE([features.h])
+AC_CHECK_HEADERS_ONCE([ifaddrs.h]) # sqUnixSocket.c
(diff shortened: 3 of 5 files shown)


dcstes:
Hash: SHA256


Any feedback please ?

The change to use <ifaddrs.h> only if the build target has it,
seems a logical change ...

The only risk is perhaps that somebody unintentionally re-enables the 
'old code' by not defining HAVE_IFADDRS_H in their Makefile or config.h.

This is because by default

#undef HAVE_IFADDRS_H

and that will thus by default enable the 'old code'.

While the default now is to enable the 'new code' which uses <ifaddrs.h>.

But this can easily be solved by requiring to #define HAVE_IFADDRS_H,
and the 'configure' script will do just that automatically if it finds the file.

Anyway -- with this patch the current COGVM compiles on Solaris 10.

By the way Solaris 10 still uses openssl 0.9.7 and I think the configure
quite correctly disables SSL support so that works fine.

The SSL plugin is not built I think on Solaris 10, but no problem.

The real interest is in the support for Solaris 11 with openssl 1.0.2 etc.,
but it is nice that the code still compiles on Solaris 10.

Regards,
David Stes

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJfbZoGAAoJEAwpOKXMq1MaO3sH/26aTh31X//H3xC/bQn+jPdK
ES/YFfFXJ/OB9i/WroLia8n4IHpoPCguwdLy5NPtMbo30s8VpQeVWe2OAun+jLOX
Ekj9LbfFbKpAftetgD7O+2ONWmbikvx1nrPcl4qfkD1rcf1m+ESbftsf69vDvIrc
rNC+Bgq4ZR84yQIxOySzZskN34mtPQfdOxw2B1JQP1cTtPvynkiBy+fGFM2m8Y3J
7StRLKQx9q4dzI875xV3Vmd9gmhFxA4MNHMOmh1i4r9rZYbrQjCtaoHnuI9abQcf
jPKI01K11kyj80aZlV3h9oaslTJnEIGWhJkHFYjQBamVJcqbj2wvYLCiwc9WHIY=
=lu3K
-----END PGP SIGNATURE-----

PR #528, opened 2020-10-18
Added epoll support for aio.c on unix

smalltalking:
When HAVE_CONFIG_H and HAVE_EPOLL is true (the former is pretty much always
true), the select-based aio implementation is replaced with an epoll-based one.
The epoll variant has a few advantages:

- it is free of the lost readSemaphore signals bug (see the comments of
  Socket's various #waitFor* methods which wait on readSemaphore).
- idle CPU usage is down from ~10% to ~2%
- faster Socket operations

Known issue:
- when the VM quits, an error message appears from aioDisable() because
  the file descriptor about to be removed from the epoll set has already
  been closed, therefore it's not in the set. It can be worked around
  by supressing error messages when aioDisable() is called by aioFini()

### DIFF
--- a/platforms/unix/vm/aio.c
+++ b/platforms/unix/vm/aio.c
@@ -33,6 +33,7 @@
 #include "interp.h" /* For COGVM define */
 #include "sqaio.h"
 #include "sqAssert.h"
+#include <stdlib.h>
 
 #ifdef HAVE_CONFIG_H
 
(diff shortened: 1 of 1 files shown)


eliotmiranda:
Wow Levente, thank you!!!!

_,,,^..^,,,_ (phone)

eliotmiranda (approved):
This is fabulous and long desired.  While the builds are failing they're doing so because of an hack for image file access in sq.h that has to go.  So I'm going to approve and move the file access macro crap to a set of platform-specific files included by the interpreter.  i.e. we're going to break things until we fix them.

PR #543, opened 2020-12-24
More manual cmake settings management

bencoman:
The previous WIP PR was copying the template file to "CMakeSettings.jsonx" (notice the x on the end).
When changed to copy over actual user "CMakeSettings.json", Visual Studio was automatically re-parsing "CMakeSettings.json" and the info messages didn't get shown long enough to understand what was happening. 

So I believe copying the template over the user build settings will need to be a manual operation per this PR.


bencoman:
whoops

PR #547, opened 2021-01-18
Fix missing KeyRelease events when multiple keys are depressed.

dtlewis290:
Reference Mantis 0007597 http://bugs.squeak.org/view.php?id=7597.
Rather than keep a single lastKey to remember the last previously
pressed key value, maintain an array size 256 of last key pressed values
indexed by X11 KeyCode. Works for any number of simultaneous keys.


eliotmiranda (commented):
Hi David, I would bite the bullet and write an explicit initializer for lastKeyValue containing 256 -1's, get rid of the initialization lines in storeLastKeyValue (lines 1722 through 1727).  Looks perfect otherwise.

dtlewis290:
OK I will do an update. Also as reminded by Tim, the index into array should probably be masked to ensure 1 -255. A KeyCode is always in that range, but there is no real guarantee that there is not some scenario in which it might be set to -1 for example, so better to be safe.

Issue #548, opened 2021-01-25
UnixProcess forkSqueak broken since October


dtlewis290:
UnixProcess class>>forkSqueak is no longer working. The forked child process VM crashes with segmentation fault. Testing with VMs from bintray shows that version 5.0-202009300634 works, and any version 5.0-202010192227 or later fails. Stack dump sometimes (but not always) shows failure in aioPoll() for example:

   */usr/local/bin/../lib/squeak/5.0-202101160259/squeak(aioPoll+0x12e)[0x4bc0fe]

I am not able to catch the failure in gdb because it happens in the child process. My initial guess is that it may be related to the epoll enhancements added in this time frame, because forking the VM requires initializing things like this in the new child VM process.

dtlewis290:
Thank you Jan!

dtlewis290:
The segfault happens in the child process that was forked by the forkSqueak prim. It occurs in the new epoll code. I don't yet see the cause (there is no obvious null pointer issue) but the gdb backtrace is:

(gdb) bt
#0  0x0000000000dd39b0 in ?? ()
#1  0x00000000004d62dc in aioPoll (microSeconds=0) at /home/lewis/squeak/git/opensmalltalk-vm/platforms/unix/vm/aio.c:405
#2  0x00007fa016315e39 in display_ioProcessEvents () at /home/lewis/squeak/git/opensmalltalk-vm/platforms/unix/vm-display-X11/sqUnixX11.c:4867
#3  0x0000000000417ca3 in ioProcessEvents () at /home/lewis/squeak/git/opensmalltalk-vm/platforms/unix/vm/sqUnixMain.c:726
#4  0x0000000000441f58 in checkForEventsMayContextSwitch (mayContextSwitch=1) at /home/lewis/squeak/git/opensmalltalk-vm/spurstack64src/vm/gcc3x-interp.c:50306
#5  0x00000000004401ca in handleStackOverflowOrEventAllowContextSwitch (mayContextSwitch=1)
    at /home/lewis/squeak/git/opensmalltalk-vm/spurstack64src/vm/gcc3x-interp.c:53718
#6  0x0000000000426cbd in interpret () at /home/lewis/squeak/git/opensmalltalk-vm/spurstack64src/vm/gcc3x-interp.c:5844
[...]

PR #550, opened 2021-02-01
In forgetXDisplay() do not call aioDisable() for the socket connection

dtlewis290:
to the X11 server if Linux epoll event handling is being used, because
forgetXDisplay() is typically called after a fork(). With epoll event
handling the shared fd reference affects both parent and child badly.

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -6797,8 +6797,10 @@ int forgetXDisplay(void)
   stDisplay= null;      /* Squeak display                       */
   if (isConnectedToXServer)
     close(stXfd);
+#if ! (HAVE_CONFIG_H && HAVE_EPOLL)
   if (stXfd >= 0)
     aioDisable(stXfd);
+#endif
   stXfd= -1;		/* X connection file descriptor         */
   stParent= null;
   stWindow= null;       /* Squeak window                        */


nicolas-cellier-aka-nice:
IMO, not a good fix.
The fact the the epoll structure is shared between parent and childs (forked) process is a problem not only for the X11 socket connection, but for ALL the polled fd.
We shall close the epoll fd and re-create it in the forked vm.
Probably, we should use a close-on-exec flag for epoll `epoll_create1(EPOLL_CLOEXEC )`, so as to protect from accidental sharing, unless someone foresee a good scenario for sharing epoll fd (I fail to see one).

krono:
Isn't there a way to duplicate fd's on fork, so that the child can close them at will without harm?
(haven't done fork/exec in a while, tho)

nicolas-cellier-aka-nice:
> 
> 
> Isn't there a way to duplicate fd's on fork, so that the child can close them at will without harm?
> (haven't done fork/exec in a while, tho)

No, from what I understand, the file-descriptors are duped anyway (they are in the process space), but the file descriptions are shared (in the kernel space). The duped fd point to shared objects...
See https://copyconstruct.medium.com/the-method-to-epolls-madness-d9d2d6378642

krono:
Regarding `EPOLL_CLOEXEC`, does it really help when we do not actually `exec`?

(Edit: reading [this](https://idea.popcount.org/2017-03-20-epoll-is-fundamentally-broken-22/) suggests that we `epoll_ctl(EPOLL_CTL_DEL)` before we `close()` whenever we have the fd `epoll`d)

nicolas-cellier-aka-nice:
Ouch, you're right, close-on-exec does close on exec as the name tells...
Why the hell did I think that it would work at fork time? Just because it would fit our needs? Sort of wishful thinking...

dtlewis290:
I made a couple of commits to the Cog branch that provide error messages on stderr rather than segfault the VM. This shows clearly that after a fork, the X11 event notification for the child process is being delivered to both the child VM and the parent VM, and also that the kernel is supplying empty event data to the parent when this happens.

dtlewis290:
I suspect that for epoll, the rule should be never unregister the fd if you are also going to close the fd. This would be different from non-epoll event event handling, so possibly there would be a need for e.g. aioDisableAndClose(fd) in addition to the existing aioDisable(fd). Or maybe is it sufficient to just reopen the aiofd after a fork as suggested above.

dtlewis290:
Closing the PR because the solution is not sufficient. I will open a different PR to close and reopen the epoll fd based on the suggestions above and in issue #548.

PR #552, opened 2021-02-08
For epoll aio, close and reopen the epoll fd in forked child process

dtlewis290:
Factor the epoll initiation out of aioInit into new epollInit function
and invoke it via pthread_atfork in the child process after a fork.
Prevents misdelivery of events from the common kernel epoll structures.


smalltalking:
I see no problems with this. Merge it.

PR #553, opened 2021-03-08
OS X custom URL scheme and document open event injection fixup

ronsaldo:
These change fixup the events for handling custom url schemes, and the support for opening document types registered in the plist that are received as events before the image is even opened by the VM. These event are injected as drop event into the image for further processing.


eliotmiranda (approved):
Best to get this in and fix any problems in situ.  I want this :-)

PR #563, opened 2021-04-30
Fix UDP binding in Windows sockets

jdougan:
Changed sqSocketListenOnPortBacklogSizeInterface() to match general behavior of Unix socket version. Previously was always incorrectly binding UDP sockets to INADDR_ANY instead of the specified IPv4 address.

### DIFF
--- a/platforms/win32/plugins/SocketPlugin/sqWin32NewNet.c
+++ b/platforms/win32/plugins/SocketPlugin/sqWin32NewNet.c
@@ -993,7 +993,7 @@ void sqSocketListenOnPortBacklogSizeInterface(SocketPtr s, sqInt port, sqInt bac
   if (!SocketValid(s)) return;
 
   if (TCPSocketType != s->socketType) { /* UDP/RAW */
-	sqSocketListenOnPort(s, port);
+	sqSocketBindToPort(s, (int)addr, (int)port);
 	return;
   }
 


eliotmiranda:
Hi John, thanks for this.  Which specific tests are affected by this change?

jdougan:
No tests present in 5.3 broke. I could redo my test code and stick it in the socket tests, if you could point me to how to get it integrated.  I caught it while working on Rbt to make it handle multiple network interfaces more sanely.

This prior behavior was normal up until Sept. 2003, when there was discussion on the squeak-l about it and Ian updated the Unix code to do the right thing. Since then there have been some complaints, but apparently no one who had the expertise or time to fix it was available.  Is there anyone at present who has the nominal responsibility for the socket plugin and the various platform modules? I spent a lot of time reading the code and have some questions about the whys of the design.

eliotmiranda:
John, look at SocketTest in a 5.3 or preferably a trunk image.  You can then submit a new version of the Tests package to inbox or just mail me a change set etc. We can discuss more on squeak-dev to eg make you a committer etc.

Discuss the design issues on vm-dev or squeak-dev.  I’m curious about the design too and would love to get rid of the threads and the asynchronous name lookup, replaced by a proper threaded FFI.

smalltalking:
I noted this issue along with some others back in 2014 but never got my questions answered: http://lists.squeakfoundation.org/pipermail/vm-dev/2014-December/017383.html
SocketPlugin's platform code needs some review/cleanup.

PR #573, opened 2021-07-16
Move all flavors of CogVM sources to src/vm.*; move all build.* flavors to building/*

marceltaeumel:
Moves all flavors of CogVM sources into "src/" folder.
http://lists.squeakfoundation.org/pipermail/vm-dev/2021-July/036540.html

Moves all build folders down to "building/" folder. Chose "building" over "build" to avoid conflicts with common build systems.
http://lists.squeakfoundation.org/pipermail/vm-dev/2021-July/036551.html

For updates to VMMaker, see `VMMaker.oscog-mt.2987` in VMMakerInbox.

Note that this PR was successfully tests through our GitHub Actions workflows: https://github.com/OpenSmalltalk/opensmalltalk-vm/actions

## How to Review This

This PR shows over 3 million lines of text added and about the same number removed. This is not how one should review this PR. Instead, go through the list of commits and take a look at each message and diff. While doing so, you should ignore the following commits because those are either a "move files" operation or a "update CogVM sources":

- Move generated sources from "\*src/vm" to "src/vm.\*": 5d32f6d 0ebb846
- Move build files from "build.\*" to "building/\*": 0adda86 
- Update CogVM sources to avoid conflicts with upstream: 95f1068 255b3d2 de84852 f1c0a95
- Various merge-commits from upstream: 0ac5379 c488ac4 5796836

All other commits have rather simple diffs, accompanied with explanatory commit messages.


eliotmiranda:
Since v3 is implicitly 32-bits how say we
- lose the .32bit & .64bit and just have spur32 & spur64?
- lose vm; it's also implicit given the plugins directory
So then we'd have
src/
ckformat.c
examplePlugins.ext
examplePlugins.int
lowcode.cog.spur32
lowcode.stack.spur32
lowcode.cog.spur64
lowcode.stack.spur64
newspeak.cog.spur32
newspeak.stack.spur32
newspeak.cog.spur64
newspeak.stack.spur64
plugins
cog.spur32
cog.v3
sista.spur32
stack.spur32
stack.v3
cog.spur64
sista.spur64
stack.spur64

And to my mind the object memory architecture is primary (since it determines image compatibility), so I would go for
ckformat.c
examplePlugins.ext
examplePlugins.int
plugins
spur32.cog
spur32.cog.lowcode
spur32.cog.newspeak
spur32.sista
spur32.stack
spur32.stack.lowcode
spur32.stack.newspeak
spur64.cog
spur64.cog.lowcode
spur64.cog.newspeak
spur64.sista
spur64.stack
spur64.stack.lowcode
spur64.stack.newspeak
v3.cog
v3.stack

That makes *much* more sense to me.

eliotmiranda:
Also, all generated files fall foul of scripts/findUnofficialFiles (they are all generated from a dirty VM package).  This is a no no :-).  BTW, we should merely *move* the generated files from the previous commit into their new positions, not regenerate them.  e.g. the fact that spur64src/vm/cogit.c has not been regenerated since VMMaker.oscog-eem.2859 is meaningful (if only to me). i.e. I try hard to regenerate only when the generated code actually changes, not just when VMMaker changes.

marceltaeumel:
> Since v3 is implicitly 32-bits [...]

I made this explicit so that newcomers can see it immediately. No need to hide this information. :-) It also helped me writing the GitHub Actions workflows. There, 32-bit and 64-bit is kind of the primary discriminator. v3 and spur come secondly.

> And to my mind the object memory architecture is primary (since it determines image compatibility),

Hmm... I found it helpful to not drop the "vm." prefix and to move both "newspeak" and "lowcode" to the front to indicate that both are flavors not directly connected to the basic osvm development.

> Also, all generated files fall foul of scripts/findUnofficialFiles (they are all generated from a dirty VM package). This is a no no :-). 

Is there a way to read up on this? I tried to not commit comment-only changes. And because of the "Fix config.h include order"-issue, I got into the habit of re-generating the sources myself to see how good we can use the GitHub-Tooling for such things. :-)

eliotmiranda:
OK, ciao bella.

marceltaeumel:
Hmm... so you did not like the path structure in "src" but merged the PR anyway? I am confused. Well, it's very simple to change the structure as you proposed. I thought I double-checked that with you on vm-dev weeks ago.

marceltaeumel:
In PR #587, I just restored the CogVM sources to what you committed in https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/12b0d50876312a74c51be3a81232b55a4b3efb7f. See https://github.com/marceltaeumel/opensmalltalk-vm/commit/3254c5460e29e430c3f86502e2a35ee9c8e2e775

Issue #575, opened 2021-07-25
Unable to rebind keyboard device in vm-display-fbdev (FrameBuffer)


KenDickey:
With libevdev, mouse and keyboard are bound to devices in /dev/input with names like "event0" and "event1".
Different keyboard/mouse hw binds to different numbers, so it is important to allow overrides via shell variables.
E.g. 
SQUEAK_MSDEV=/dev/input/event1
SQUEAK_KBDEV=/dev/input/event0
vs 
SQUEAK_MSDEV=/dev/input/event0
SQUEAK_KBDEV=/dev/input/event1

Mouse rebinding works, but not keyboard rebinding.

KenDickey:
Using evtest I was able to determine device mapping
 /dev/input/event0 -> menu-keys
    event1 -> audio jack
    event2 -> track pad
    event3 -> mouse
    event4 -> keyboard (raw keystrokes)
    event5 -> system control (e.g. sleep)
    event6 -> keyboard events (keycodes)

So binding 
  SQUEAK_KBDEV=/dev/input/event6
  SQUEAK_MSDEV=/dev/input/event3
then the Framebuffer display works just fine

PR #577, opened 2021-07-27
Fixes include order for ARM32 builds. Thanks to Bruce O'Neel!

marceltaeumel:


KenDickey:
Verified clean build with build.linux32ARMv6/squeak.{stack,cog}.spur/build/mvm
on RaspBerry Pi 3 running Raspberry Pi OS.

Note that I can't test either stack or cog VMs as no X11 and vm-display-fbdev needs pulll/merge to fix issue #575.  But the VM binaries build just fine,  [Also, VectorEnginePlugin not yet integrated].  FYI.

PR #578, opened 2021-07-29
Add vector engine plugin

dtlewis290:
Update the VMMaker build scripts to add a repository for VectorEnginePlugin, using a squeaksource mirror of Juan's repository for convenience. Development source for VectorEnginePlugin is Packages/Features/VectorEnginePlugin.pck.st from repository https://github.com/Cuis-Smalltalk/Cuis-Smalltalk-Dev.

Disable compiler optimization for VectorEnginePlugin. For gcc this is not an issue, but with llvm any compiler optimization
results in a broken plugin (various symptoms). Add -O0 to XCFLAGS in makefile.inc to override any previous definition of -O in CFLAGS.

Not included in this PR: VectorEnginePlugin needs to be added to various plugin.ext files. So far I have tested only X86 Linux, so this should be the subject of separate updates.


dtlewis290:
I went ahead an added the generated source VectorEnginePlugin.c and updated the plugins.ext for most Linux configurations.

dtlewis290:
As far as I can tell this is a safe update. If no objections, I'll merge it a couple of days from now.

KenDickey:
Juan Vuletich reports testing with VectorEnginePLugin:

"I've tried it on MacOS, Ubuntu X64 and Windows X64, running on a fast I9 MacBook. Also tried Dave's 32 bit interpreter for Linux X64 on the same machine. I also tried Win 32 on a 2010 Atom netbook. All them run fine."

Looks good to integrate into "platforms/Cross/plugins"

dtlewis290:
Prompted by notes from Marcel, I realize that I made a mistake in the patch I did for the VMMaker build scripts. I want to withdraw those update until I can check with Eliot as to how best to handle Cuis underscore assignment characters when loading the plugin from either Cuis GitHub or the separate squeaksource mirror of the Cuis GitHub repository. For this reason I am going to close this pull request. I will cherry-pick the actual C source commits so that these will be available for general use.

KenDickey:
Confirm  current is working with:

build.linux64ARMv8/squeak.stack.spur/build/mvm clean on Aarch64 Chromebook 2
  and
build.linux32ARMv6/squeak.cog.spur/build/mvm clean on Raspberry Pi 3.

Thanks much!!
-KenD

PR #579, opened 2021-07-30
Recover automated builds via GitHub Actions

marceltaeumel:
Adds workflows (i.e., GitHub Actions) for automated builds. Replaces TravisCI and AppVeyor integration.

Automated builds on push/pull-requests:
 - Linux (x86, 64-bit and 32-bit)
 - Linux (ARMv6, ARMv7, ARMv8/Aarch64)
 - macOS (x86, only 64-bit)
 - Windows (x86, 64-bit and 32-bit)

Mostly, those builds include the following flavors:
 - squeak.cog.spur (also for Cuis)
 - squeak.sista.spur (also for Cuis)
 - squeak.stack.spur (also for Cuis)
 - pharo.cog.spur (for Pharo7.0 and before, I suppose)
 - pharo.sista.spur (for Pharo7.0 and before, I suppose)
 - pharo.stack.spur (for Pharo7.0 and before, I suppose)

There are some exceptions from this pattern such as:
 - pharo.* not building on Windows bc. of an issue with third-party lib builds, related to GitHub Actions
 - *.sista.spur bc. of unprepared build target on some platforms such as all ARM targets
 - No Linux itimer-heartbeat builds prepared for *.stack.spur; threaded only for now
 - No pharo.* builds prepared for ARM so far, except pharo.cog.cpur for ARMv6 in monthly builds; see below

Still, many if not all of those issues can easily be fixed in the future. We would just add the missing build configurations. :-)

Note that all these VM flavors include builds for "fast," "assert," and "debug" mode -- which we did not have in our previous TravisCI/AppVeyor setup.

---------------

There are also two separate workflows that manage "extra builds," such as for VM flavors that are not used on a regular basis:
 - newspeak.* (incl. ARMv7)
 - *.v3 (incl. ARMv6)
 - *.lowcode
 - pharo.cog.spur for ARMv6

These extra builds are triggered *once per month* or anytime per request.

These extra builds are only in "fast" mode.

---------------

All the latest build artifacts can be downloaded from three different tags, which are for my fork as follows:
 - https://github.com/marceltaeumel/opensmalltalk-vm/releases/tag/latest-build
 - https://github.com/marceltaeumel/opensmalltalk-vm/releases/tag/latest-assert-build
 - https://github.com/marceltaeumel/opensmalltalk-vm/releases/tag/latest-debug-build

After merging this pull-request, replace "marceltaeumel" with "opensmalltalk-vm", and you will form the respective URLs. The URL will be stable for any particular VM flavor and platform. **Feel free to share those URLs to "Get the latest bleeding edge OSVM now!"** :-) 

For previous build artifacts, note that all workflows store their artifacts *with a timestamp* after they ran. You can access those artifacts for about 90 days when logged into GitHub. Here is an example:
 - https://github.com/marceltaeumel/opensmalltalk-vm/actions/runs/1073960213

Note that you can download a ZIP archive from GitHub, which will then hold the actual build artifact/container, e.g., \*.zip or \*.tar.gz or \*.dmg.

---------------

Finally, we can now manage builds and artifacts for release candidates:
 1. Tag a commit as candidate such as "202107301330"
 2. For each "Build for *"-workflow on the actions page for OSVM, just click "Run workflow" where you enter that tag and manually start the workflow.
 3. Those artifacts will be attached to the tag and are then accessible under, e.g., "https://github.com/marceltaeumel/opensmalltalk-vm/releases/tag/202107301330", just like the *latest-build* artifacts are.

For an actual VM release, we would manually create a release through the GitHub webpage and copy over the artifacts from the latest release candidate. For the squeak-app bundles, we would also copy those over to http://files.squeak.org/base/.

---------------

After merging this PR, an *owner* in this repository should:
 - Update the certificates in deploy/squeak, deploy/newspeak, and deploy/pharo and configure the required secret as documented in .github/workflows/macos.yml
 - Deactivate the TravisCI and AppVeyor hooks. Their .yml files have been removed already.

Happy coding!


eliotmiranda:
Hi Marcel,

    I think it’s safe to nuke the itinerary builds on all Linux platforms now.  The itinerary builds are a hack to get around the inability to define a heartbeat thread that runs at higher priority then the Smalltalk thread, which is a limitation of pre 2.6 kernels.  So this is ancient history we can say good bye to.  Unless a Linux expert says otherwise n the next day I’d two I’ll nuke all the itimer support.

_,,,^..^,,,_ (phone)

marceltaeumel:
Yes, if there is no need for the "itimer"-heartbeat versions, I can remove those from the build jobs. Yet, I would wait for some active Linux users to verify this assumption.

fniephaus (approved):
LGTM. Thanks for the work, @marceltaeumel!

nicolas-cellier-aka-nice:
Great work, thank you!
Concerning itimer, I did use it so as to compile/run on WSL.

PR #587, opened 2021-08-25
Updates src paths to follow Eliot's suggestions

marceltaeumel:
- See https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/573#issuecomment-905099419
- See VMMaker.oscog-mt.3049 in VMMakerInbox


eliotmiranda:
thank you!

PR #590, opened 2021-09-05
Use distinct VERSION_TAG based on sizeof(void *) to distinguish 32/64…

dtlewis290:
… bit

installation libraries. VERSION_TAG is used for naming the installation
directory, append "-32bit" or -"64bit" to VERSION_TAG to enable 32-bit
and 64-bit VMs to be installed side by side in standard unix directories.

### DIFF
--- a/platforms/Cross/vm/sqSCCSVersion.h
+++ b/platforms/Cross/vm/sqSCCSVersion.h
@@ -239,7 +239,7 @@ main(int _argc, char **_argv)
 	if (printit("VM_RELEASE"))
 		printf("%s\n", revisionAsString());
 	if (printit("VERSION_TAG"))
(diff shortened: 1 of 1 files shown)


dtlewis290:
Closed because the ./products build contain a "squeak" script that refers to the now-renamed ./bin script. In a normal "make install" in which one installs in default unix directories, this "squeak" script is (improperly) installed in /usr/local/ where it is functionally harmless. However, in the products build the script needs to work. I'll resubmit a PR if and when I can find a workaround for that issue.

dtlewis290:
Reopening. I retested after cleaning out my ./products directory, and all is working as expected, no issues with builds in ./products, and standard unix install to e.g. /usr/local/bin/ works. I think this PR can be safely merged to Cog but a second set of eyes on it would be appreciated.

PR #591, opened 2021-09-06
Update mvm scripts to specify distinct names for squeak run scripts.

dtlewis290:
sudo make install (following build with ./mvm) will install 64-bit Spur VMs
with /usr/local/bin/spur64, 32-bit Spur VMs with /usr/local/bin/spur, and
32-bit Cog V3 VMs as /usr/local/bin/cog. Along with branch dtl/distinct-VERSION_TAG
(separate PR), this allows the three flavors of opensmalltalk-vm, plus the
interpreter VM, to all be installed side by side in a standard unix directory
(e.g. /usr/local/bin/ or /usr/bin for executable scripts).  Within the various
builds in ./products, the base squeak script is still called "squeak", and it
invokes ./bin/spur64, ./bin/spur or ./bin/cog in that folder depending on
the type of VM. Net effect is no change in behavior for the ./products builds,
and elimination of name space conflicts in the standard unix directories.


marceltaeumel:
It looks like you also need to update "configure" and "configure.ac"? Why is it "spur" and not "spur32" but "spur64"? Is this a common pattern on unix/linux?

dtlewis290:
The naming convention is historical, just what I have been using on my system for some years. First we had /usr/bin/squeak, then Cog became available, so /usr/bin/cog. Then Spur was added, so /usr/bin/spur. Later the 64-bit Spur was added, so /usr/bin/spur64. At this point it might be more clear to use the name spur32 rather than spur?

dtlewis290:
The updates for configure and make are in https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/a385f5b69e7bda720e980dd61e683a0bf4810cd7 in the Cog branch which should be sufficient to support this PR.

eliotmiranda:
On Fri, Sep 10, 2021 at 8:55 AM David T Lewis ***@***.***>
wrote:

> The naming convention is historical, just what I have been using on my
> system for some years. First we had /usr/bin/squeak, then Cog became
> available, so /usr/bin/cog. Then Spur was added, so /usr/bin/spur. Later
> the 64-bit Spur was added, so /usr/bin/spur64. At this point it might be
> more clear to use the name spur32 rather than spur?
>

I think so.  It's a nice symmetrical distinction.  On Unix /usr/lib is
ambiguous.  Some systems added /usr/lib64, some /usr/lib32,, all in an
effort to not confront the necessity of changing the old /usr/lib into
/usr/lib32.  We don't have to repeat the same mistake/  It's a litt;e
painful at fiurst but using spur32 & spur64, and not using spur is clear
and easier for scripts, etc.  IMO.
_,,,^..^,,,_
best, Eliot

dtlewis290:
Agreed, I just pushed an update to use spur32 as script name for 32-bit Spur VMs.

marceltaeumel:
Still not working:
```
configure: error: unrecognized option: `--scriptname=spur32'
```
For example, see https://github.com/OpenSmalltalk/opensmalltalk-vm/pull/591/checks?check_run_id=3620475617

`make configure` in `platforms/unix/config` seems to forget that option again. Hmm....

dtlewis290:
Is it possible that the CI is failing to update the platforms/unix/config directory? The "missing" configure option is present in the platforms sources but seems to be missing when run on CI.

marceltaeumel:
I will push a fix for configure.ac to this PR soon. It's almost ready. :-) I will change the option "scriptname" to "with-scriptname" because it is then compatible with autoconf.

marceltaeumel:
Done. Please check if I broke anything. "sudo make install" still works as expected in my VirtualBox. 😄

dtlewis290:
Aha, I see it now, I was not doing a make on the config directory so did not notice the conflict. Thank you for the fix Marcel.

dtlewis290:
Confirming - the builds for spur64, spur32, and cog all work as expected when building on my Ubuntu laptop. Correct scripts are installed in /usr/local/bin and the ./products folder looks good also.

PR #592, opened 2021-09-06
Update BitBltArmSimdSourceWord.s to latest ben avison version

timrowledge:
Half a dozen small bug fixes from ben

Co-Authored-By: bavison <3324657+bavison@users.noreply.github.com>


krono:
@timrowledge does this somehow complement #565 ?

timrowledge:
Wow, this is from a while ago. I'd have to say that yes, it definitely relates to #565.
I'm afraid I have no idea what the error message means and qick googling helps not at all. Best guess is some type stuff mismatch but I do know for certain it did compile originally.

krono:
@marceltaeumel opinion?

PR #593, opened 2021-09-07
`WindowEvent`[`A`|`Dea`]`ctivated`

LinqLover:
This PR implements the recording of `WindowEventActivated` on all primary platforms (that is, Win32 and Linux; apple does not count unless they provide adequate virtualization support for other platforms). In the past, this event was only declared but never recorded. In addition, the new event `WindowEventDeactivated` is declared and recorded, too.

A real application example for this is proposed on the squeak-dev list [here](http://lists.squeakfoundation.org/pipermail/squeak-dev/2021-November/216931.html), but I do not think that the discussion about this proposal should hinder us from integrating this change into the VM, should it?


LinqLover:
To whomever it may concern: This is ready for review now! :-)

eliotmiranda (approved):
Looks good, albeit with the conspicuous absence of a Mac implementation.  Approving, and hopefully someone will contribute the Mac code soon.

LinqLover:
Thank you Eliot! :-)

PR #595, opened 2021-09-17
Adds action for building squeak.stack.spur on Linux using the generic…

marceltaeumel:
… mvm files, which are prepared for our Debian builds. No update of latest-build release but artifact upload for 90-day reviews. Fixes minor issue in configure.ac to be able to build 32-bit binaries on x86_64 platforms again.


krono:
@marceltaeumel is this still relevant?

PR #599, opened 2021-10-01
Adds cross-platform compatible virtual-key codes to vm-display-x11 plugin

marceltaeumel:
Since X11 does not provide a localized version of virtual keys, I mapped the X11 event's `KeySym` to such a "virtual-key code", using `evt.keycode` only as a fall back. It's not perfect but close to what Windows and macOS have.

Here is the mapping schema:

```
 * 0x00         (unknown virtual key)
 * 0x01         XK_space (shifted from Latin-1)
 * 0x02         XK_comma (shifted from Latin-1 OR guessed from xKeycode)
 * 0x03         XK_period (shifted from Latin-1 OR guessed from xKeycode)
 * 0x04 .. 0x07 (unused)
 *
 * 0x08 .. 0x1f SPECIAL KEYS 1 (from all-mask 0xff00)
 * 0x20 .. 0x29 NUMERIC ASCII KEYS (shifted from Latin-1)
 * 0x2a .. 0x43 ALPHABETIC ASCII KEYS (shifted from Latin-1)
 *
 * 0x44         US_MINUS (guessed from xKeycode)
 * 0x45         US_PLUS (guessed from xKeycode)
 * 0x46         US_LBRACKET (guessed from xKeycode)
 * 0x47         US_RBRACKET (guessed from xKeycode)
 * 0x48         US_COLON (guessed from xKeycode)
 * 0x49         US_QUOTE (guessed from xKeycode)
 * 0x4a         US_TILDE (guessed from xKeycode)
 * 0x4b         US_BACKSLASH (guessed from xKeycode)
 * 0x4c         US_COMMA (guessed from xKeycode)
 * 0x4d         US_PERIOD (guessed from xKeycode)
 * 0x4e         US_SLASH (guessed from xKeycode)
 * 0x4f         US_102 (guessed from xKeycode)
 *
 * 0x50 .. 0xff SPECIAL KEYS 2 (from all-mask 0xff00)
```

Here is the new mapping table for Squeak:
<s>[new-x11-virtual-keys.1.cs.txt](https://github.com/OpenSmalltalk/opensmalltalk-vm/files/7269113/new-x11-virtual-keys.1.cs.txt)</s>
[new-x11-virtual-keys.6.cs.txt](https://github.com/OpenSmalltalk/opensmalltalk-vm/files/7286590/new-x11-virtual-keys.6.cs.txt) (updated 2021-10-06 14:00 UTC)

See /usr/include/X11/keysymdef.h to understand the "SPECIAL KEYS" (masked 0xff00).

The biggest remaining issue is all those "OEM keys", which means the ones you push for "+" or "/" or "~" and so on.

*Note that the idea of having a localized version of the virtual key is that you can print it in an application's user interface. Would be really annoying if an app insists on, e.g., Z while you actionally pushed Y.*

You can try out the changes in the `KeyboardExerciser` in Squeak 6.0alpha. The test is that the tool shows the labels of your physical keyboard while you keep on pressing keys or key combinations. Note that you need a physical keyboard that matches your keyboard layout. That is, US keyboard for US key layout.

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -1992,9 +1992,109 @@ static int x2sqKeyPlain(XKeyEvent *xevt, KeySym *symbolic)
   return charCode;
 }
 
+/* Maps X11 key symbols to a one-byte, cross-platform compatible virtual-key code.
+ *
+ * 0x00         (unknown virtual key)
+ * 0x01         XK_space (shifted from Latin-1)
+ * 0x02         XK_comma (shifted from Latin-1 OR guessed from xKeycode)
+ * 0x03         XK_period (shifted from Latin-1 OR guessed from xKeycode)
+ * 0x04 .. 0x07 (unused)
+ *
+ * 0x08 .. 0x1f SPECIAL KEYS 1 (from all-mask 0xff00)
+ * 0x20 .. 0x29 NUMERIC ASCII KEYS (shifted from Latin-1)
+ * 0x2a .. 0x43 ALPHABETIC ASCII KEYS (shifted from Latin-1)
+ *
+ * 0x44         US_MINUS (guessed from xKeycode)
+ * 0x45         US_PLUS (guessed from xKeycode)
+ * 0x46         US_LBRACKET (guessed from xKeycode)
+ * 0x47         US_RBRACKET (guessed from xKeycode)
+ * 0x48         US_COLON (guessed from xKeycode)
+ * 0x49         US_QUOTE (guessed from xKeycode)
+ * 0x4a         US_TILDE (guessed from xKeycode)
+ * 0x4b         US_BACKSLASH (guessed from xKeycode)
+ * 0x4c         US_COMMA (guessed from xKeycode)
+ * 0x4d         US_PERIOD (guessed from xKeycode)
+ * 0x4e         US_SLASH (guessed from xKeycode)
+ * 0x4f         US_102 (guessed from xKeycode)
+ *
+ * 0x50 .. 0xff SPECIAL KEYS 2 (from all-mask 0xff00)
+ */
 static int xkey2sqVirtualKeyCode(int xKeycode, KeySym xKeysym)
 {
-  return xKeycode;
+  if (xKeysym >= XK_0 /* 0x30 */ && xKeysym <= XK_9 /* 0x39 */)
+    return xKeysym - 0x10; // to start at 0x20
+  if (xKeysym >= XK_A /* 0x41 */ && xKeysym <= XK_Z /* 0x5a */)
+    return xKeysym - 0x17; // to start at 0x2a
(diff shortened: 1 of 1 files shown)


krono:
Virtual key codes are _per se_ fine with me.
But I would really opt to stand on the shoulders of giants and not invent our own.
I think that SDL or QT might have the most compatible list
(libsdl.org is down, so no link…)

Or if we are interested in the physical ones, [USB HID Usage IDs](https://usb.org/sites/default/files/hut1_21.pdf) or SDL scancodes.

marceltaeumel:
> But I would really opt to stand on the shoulders of giants and not invent our own.

Sure. This here is a fix that does not require touching all other platforms. Since both macOS and Windows are fine and supported in Squeak, only X11 seems to be the current issue.

krono:
what keycodes are we using in the other worlds?

marceltaeumel:
> I think that SDL or QT might have the most compatible list

SDL does in C what Squeak 6.0alpha already does in Smalltalk. Looking at "SDL_x11keyboard.c" or "scancodes_windows.h", we have in `EventSensor >> virtualKeyTable`.

Now, SDL has its own `KeySymToSDLScancode` table. In this PR, I chose to simplify matters by just converting the special keys from 0xff__ to 0x00__ plus using the room between 0x20 and 0x4f for alpha-numeric keys. So that we stay in 1 byte, just as macOS and Windows do.

Also, SDL uses a similar "hack" for alpha-numeric keys. See SDL_x11keyboard.c function "X11_KeyCodeToSDLScancode".

This PR is als complete as in can get for X11. I am currently fixing up some modifier issues when pressing only modifier keys.

krono:
My issue is that we migh be introducing some new code here without having them mean anything anywhere else…

Also, what does "virtual" mean here? are we trying to map to the thing printed on the keyborard or to the position of the key pressed?

marceltaeumel:
> My issue is that we migh be introducing some new code here without having them mean anything anywhere else…

Yes. It would be better if we had access to both the "keysym" and the "keycode" in the Smalltalk image. Unfortunately, there is only a single value reserved for this in our event structure. So, we have to merge both value ranges in a meaningful way. I just did that. You can restore the "keysym" in the image by setting 0xff__ again. 

Do you have a better Idea on merging "keysym" (special keys and alpha-num latin-1) and "keycode" (for shifted alpha-num keys) into a single value? That single-byte lookup table in Squeak is quite handy ... and fast. :-) And what the other two platforms provide.

> Also, what does "virtual" mean here? are we trying to map to the thing printed on the keyborard or to the position of the key pressed?

The concept stems from the MSDN documentation. I elaborated on this topic several times on squeak-dev already:
- http://lists.squeakfoundation.org/pipermail/squeak-dev/2021-April/215203.html
- http://lists.squeakfoundation.org/pipermail/squeak-dev/2021-April/215208.html

The code reads `xkey2sqVirtualKeyCode`, which clarifies its meaning within the Smalltalk image.

krono:
@marceltaeumel do we want to tackle that again?

PR #620, opened 2022-04-01
For macOS -metal -core-graphics -opengl backends, fixes several issues concerning performance and high-resolution

marceltaeumel:
Tested with **Squeak6.0alpha** and **Cuis6.0-5069**.

Fixed issues:
- Correct image-side display bounds in high-resolution mode on start-up and via `#setNewScreenSize:` (i.e., HostWindowPlugin primitiveHostWindowSizeSet)
- File drag in high-resolution mode is now matching image-side display bounds
- File drag has now access to file paths in DragEnter, not just DragDrop (i.e., mouseUp)
- `-core-graphics` usable again
- `-opengl` does not slow down VM interpeter loop anymore (e.g., in Morphic's use of primitive 127 and 231)
- `-metal` does not slow down VM interpeter loop anymore (e.g., in Morphic's use of primitive 127 and 231)

Known issues:
- <s>**Rather frequent segfaults in macOS64ARMv8**, maybe related to JIT via `checkForEventsMayContextSwitch:` where `checkCogCompiledCodeCompactionCalledFor` is called before `ioProcessEvents` ... maybe because of new display events queued in `-core-graphics` and `-metal` backends</s>
- **Update:** I think I just did a mistake when partially re-compiling the VM. A full, clean compile after `make cleanall` yields a VM that did not crash for over an hour.

----

Known issues on macOS 12:
- No OS-supported "Low Resolution" mode; use magnifier
- No `-opengl` support
- High-resolution mode scales up hardware cursor

Known issues on macOS 11.6:
- No `-opengl` support
- High-resolution mode scales up hardware cursor

Known issues on macOS 10.15:
- High-resolution mode scales up hardware cursor

----

Known issues for `-metal` backend:
- Must use image-side buffer for composition to avoid flickering

Known issues for `-core-graphics` backend:
- Must use image-side buffer for composition to avoid flickering
- Blitting on `Display` via `Form reverse` not working bc. alpha value must be FF (255)
- Rather slow bc. only full and no clippy-based updates due to inaccessible double-buffering in CoreGraphics

Known issues for `-opengl` backend:
- No blitting support for Squeak's MVC
- No blitting support for Squeak's Morph's fast window dragging/resizing

-----

Note that Squeak6.0alpha has direct support for that image-side buffer via `#disableDeferredUpdates` preference. The support in Squeak's Morphic has been there since at least 1999. Cuis does not yet have an image-side buffer for composition but flickering is rare.

Note that `-core-graphics` could be made compatible with `Form reverse` blitting rule by using a color with alpha value (almost) 0 such as `Color gray alpha: 1/255` or `Color colorFromPixelValue: 16r01808080`. Try out via `Display fill: (0@0 corner: 200@200) rule: Form reverse fillColor: Color gray`.

Finally note that both `-metal` and `-core-graphics` now do rather nothing in primitive 231 (i.e., `forceDisplayUpdate`/ `ioForceDisplayUpdate` / `drawThelayers`) but rely on a scheduled display event to be delivered via primitive 94 (i.e., `primitiveGetNextEvent` / `ioProcessEvents` / `vmIOProcessEvents`), which is also called every 20 milliseconds on any other primitive send. See `checkForEventsMayContextSwitch` in VMMaker sources.

-----

**Big thanks to [krono](https://github.com/krono/) for helping out with this high-resolution stuff and other details on macOS!**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ platforms/unix/config/autom4te.cache/
 # except do track these files
 !/building/**/common*/*
 !/building/**/**/Makefile
+!/building/**/**/Makefile.*
 !/building/**/**/NotYetImplemented
 !/building/**/**/mvm
 !/building/**/**/plugins.*
--- a/building/macos64ARMv8/common/Makefile.flags
+++ b/building/macos64ARMv8/common/Makefile.flags
@@ -5,8 +5,8 @@
 
 XCODE:=$(shell /usr/bin/xcode-select -p)
 XCUB:=$(XCODE)/usr/bin
-#SDKsDIR:=$(XCODE)/Platforms/MacOSX.platform/Developer/SDKs
-SDKsDIR:=$(XCODE)/SDKs
+SDKsDIR:=$(XCODE)/Platforms/MacOSX.platform/Developer/SDKs
+#SDKsDIR:=$(XCODE)/SDKs
 #Build oldest installed SDK
 SDKs:=MacOSX.sdk
 ifeq ($(SDK),)
--- a/building/macos64ARMv8/common/Makefile.flags
+++ b/building/macos64ARMv8/common/Makefile.flags
@@ -5,7 +5,8 @@
 
 XCODE:=$(shell /usr/bin/xcode-select -p)
 XCUB:=$(XCODE)/usr/bin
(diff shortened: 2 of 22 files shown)


krono:
Pls rebase #621

PR #626, opened 2022-04-26
vm-display-X11: Fix scale factor detection in headless and no-display mode

### LINKED ISSUE #625: `primitiveScreenScaleFactor` broken on headless Linux/X11 VMs
For a few days, `primitiveScreenScaleFactor` fails on headless Linux/X11 VMs. Reprocase:

1. Download and extract the latest AiO bundle from today
2. ```bash
   $ cat << EOF > test.st
   TranscriptStream redirectToStdOut: true.
   Transcript showln: DisplayScreen actualScreenScaleFactor asString.
   EOF
   $ ./squeak.sh --headless "$(realpath test.st)"
   ```
3. stack overflow:
   ```
   DisplayScreen>>#actualScreenScaleFactor
   
   stack page bytes 8192 available headroom 5576 minimum unused headroom 6016
   
           (Segmentation fault)
   Aborted (core dumped)
   ```

8 days again, this still worked. See also: https://github.com/LinqLover/create-image/actions/workflows/test.yml

@marceltaeumel Does this require a fix similar to 2a0e7851d0e057b06e77e64a58e155d9c7935e6a?

PS: I discovered a still existing typo here: https://sourcegraph.com/search?q=context:global+ioSceenScaleFactor+repo:%5Egithub%5C.com/OpenSmalltalk/opensmalltalk-vm%24+&patternType=literal

LinqLover:
Closes #625.

With this PR, the following call does not cause a segfault from `primitiveScreenScaleFactor` any longer:

```
./squeak --headless SqueakTrunk.image --evaluate 'DisplayScreen actualScreenScaleFactor'
```

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11Scale.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11Scale.c
@@ -99,6 +99,9 @@ bool scale_Xftdpi_usable(void)
   /*
    * Xft.dpi is per se not physical and also not per-monitor.
    */
+  if (!stDisplay) {
+	return false; // running headless
+  }
   if (sqPreferPhysicalScale() || sqPerMonitorScale()) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


marceltaeumel:
Hmm... I don't like this approach. There are currently only 4 uses of the expression`!stDisplay` in "platforms/unix" and they all just verify initialization, not that headless mode. So, this is the wrong "concept" to use here.

Instead, I will go with the already used (but partially missing) `headless || !isConnectedToXServer` checks as suggested in #625.

marceltaeumel:
Superseded by https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/2d7105db755928fd90823d264fc64e17b1c57ad4

PR #635, opened 2022-05-25
[Unix/Sockets] Fix Resolution of FQDNs >64 chars

### LINKED ISSUE #619: Cannot resolve domain names longer than 64 characters on linux
SocketPlugin uses a buffer of size MAXHOSTNAMELEN + 1 on [unix](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/6d38f13347be1031168b8b02e8ca52dc1b74be1c/platforms/unix/plugins/SocketPlugin/sqUnixSocket.c#L212).
The value of that constant is 64 on linux. A domain name can have up to 255 octets with each label having up to 63 octets.
So, 64 is definitely not the right value, nor is MAXHOSTNAMELEN, because the buffer is there to store a domain name, not a host name.

krono:
We used to use MAXHOSTNAMELEN, but that is 64 on most Linuxen and
might be differen w/ or w/o POSIX compliance. We now use the
definitions from and implications by DNS itself.

Now, we can

```
WebClient httpGet: 'http://theofficialabsolutelongestdomainnameregisteredontheworldwideweb.international/'
```

Fixes #619

### DIFF
--- a/platforms/unix/plugins/SocketPlugin/sqUnixSocket.c
+++ b/platforms/unix/plugins/SocketPlugin/sqUnixSocket.c
@@ -103,10 +103,21 @@
 
 #endif /* !ACORN */
 
-/* Solaris sometimes fails to define this in netdb.h */
-#ifndef  MAXHOSTNAMELEN
-# define MAXHOSTNAMELEN	256
-#endif
+
+/*
+   As per DNS, a full DNS name cannot contain more than 255
+   characters. Due to the way this is encoded, we end up with
+   practically 253 usable bytes for any DNS FQDN out there; ther
+   cannot be a longer one.
+
+   Note: this used ot be MAXHOSTNAMELEN. Under POSIX this might be
+   aliased to HOST_NAME_MAX (or not) and/or max out to 64 characters
+   (or not). However, that would be the max length of a DNS _Label_,
+   that is, a part of a DNS name not containing any dots. This is too
+   limiting. Therefore, we define FQDN_LEN ourselves now.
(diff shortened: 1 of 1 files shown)


eliotmiranda:
The only change I would make would be to add a comment at the end of #define FQDN_LEN 253 to have it read
#define FQDN_LEN 253 // FullyQualifiedDomainName
or put it in the comment.  I had to think for a few seconds before I worked out what FQDN stood for.  Take away that pain :-)

krono:
Sorry :D
I figure(d) that fqdn is a very well known concept around DNS.
Maybe it is sufficient to expand FQDn in the comment above it?

eliotmiranda:
> On May 26, 2022, at 1:09 PM, Tobias Pape ***@***.***> wrote:
> 
> ﻿
> Sorry :D
> I figure(d) that fqdn is a very well known concept around DNS.
> Maybe it is sufficient to expand FQDn in the comment above it?
> 

Yes, that will work :-)

philipschilling:
It's actually not directly related to this issue but if you need another long domain for testing, you can try [https://taumatawhakatangihanga.koauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu.nz](https://taumatawhakatangihanga.koauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu.nz)

Interestingly, the SSL certificates of my web provider also supported only 64 characters for the full domain but in some cases it can be much longer. The domain above has 89 characters.

PR #644, opened 2022-07-21
Riscv64/rv64gcv

KenDickey:
Both X11 and Framebuffer displays work with Cuis and Squeak images on Debian Linux.
Cuis FFI and BaseImage unit tests all green.
Good time to integrate w current before things get out of sync due to high commit rate to opensmalltalk-vm git base.

Notes:
-  LicheeRV Dock HW notes and VMMaker delta St files in 
  /building/linux64riscv
- Generated /src/plugins/SqueakFFIPrims/RiscV64FFIPlugin.c has been hand altered from VMMaker generated code to split float & double return callouts as I lack the deep knowledge to setup the code generation for ffiCallArgArrayOrNilNumArgs().  Code runs fine as altered.
- While base images run fine, including Vector Graphics, there is a CRC error when reading PNG files which is under investigation.  On balance, I would prefer to integrate now and handle this separately.


krono:
This looks very nice! Thank you

eliotmiranda (approved):

PR #646, opened 2022-07-25
Refixed slip in mvm

KenDickey:
Elided 1 line (TARGET_ARCH)


KenDickey:
More bits & bobs for RV64g.  All tested on riscv64.

eliotmiranda (changes requested):
Hi Ken,  can you do a different version which doesn't include either of
    src/plugins/SqueakFFIPrims/RiscV64FFIPlugin.c
    src/plugins/ZipPlugin/ZipPlugin.c
?
I generate these carefully. Your generations are against changed VMMaker packages (see the static const char *moduleName = "ZipPlugin * VMMaker.oscog-eem.3218 " INT_EXT; lines).  So I don't want to accept these changes and have to regenerate the plugin files again.  Everything else looks fine.

KenDickey:
nonRegisterStructReturnIsViaImplicitFirstArgument still needs to return false/0

If you remove VMMaker method
  ThreadedARM64FFIPlugin>>nonRegisterStructReturnIsViaImplicitFirstArgument

The inherited behavior should be correct.

I am not much of a GIT user.  Is current close enough?

eliotmiranda (approved):

PR #648, opened 2022-08-02
Works w current vm + Alien-Core-KenD-124.mcz

KenDickey:
riscv64abicc.c for AlienInbox:Alien-Core-Kend-124

### DIFF
--- a/platforms/unix/config/configure
+++ b/platforms/unix/config/configure
@@ -12093,7 +12093,7 @@ $as_echo "$build_arch" >&6; }
 		;;
 
 		riscv64) TARGET_ARCH="-march=rv64gcv -mabi=lp64d"
-			CFLAGS="-g -O2 -D__linux__ -D__riscv64__"
+			CFLAGS="-g -O2 -D__riscv64__"
 		;;
 
 		*)
--- a/building/linux64riscv/squeak.stack.spur/build.assert/mvm
+++ b/building/linux64riscv/squeak.stack.spur/build.assert/mvm
@@ -22,8 +22,8 @@ test -f config.h || ../../../../platforms/unix/config/configure \
 	--with-scriptname=spur64 \
 	CC=gcc \
 	VM_WORD_SIZE="64" \
-	VM_CFLAGS="$OPT -D__linux__ -D__riscv64__"
-	CFLAGS="$OPT  -D__linux__ -D__riscv64__"
+	VM_CFLAGS="$OPT"
+	CFLAGS="$OPT"
 rm -f vm/sqUnixMain.o # nuke version info
 rm -rf ../../../../products/$INSTALLDIR
 # prefer make install prefix=`readlink -f \`pwd\`/../../../../products/$INSTALLDIR`
--- a/building/linux64riscv/squeak.stack.spur/build.debug/mvm
+++ b/building/linux64riscv/squeak.stack.spur/build.debug/mvm
@@ -24,8 +24,8 @@ test -f config.h || ../../../../platforms/unix/config/configure \
 	--with-scriptname=spur64 \
 	CC=gcc \
 	VM_WORD_SIZE="64" \
-	VM_CFLAGS="$OPT -D__linux__ -D__riscv64__"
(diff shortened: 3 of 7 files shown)


platforms/unix/config/configure
```diff
@@ -12093,7 +12093,7 @@ $as_echo "$build_arch" >&6; }
 		;;
 
 		riscv64) TARGET_ARCH="-march=rv64gcv -mabi=lp64d"
-			CFLAGS="-g -O2 -D__linux__ -D__riscv64__"
+			CFLAGS="-g -O2 -D__riscv64__"
```
krono, line 12096: @KenDickey Plesae, put the `configure` changes into `configure.ac` otherwise they are gone the next time autoconf runs…

KenDickey: Thanks.  I believe I have all the compiler flag fixes in.

KenDickey:
Tested 'Alien exampleCqsort' with images built from mvm's on build, build.debug, and build.assert branckes.  AOK.

platforms/unix/config/configure
eliotmiranda: Ken,  I feel uneasy with the unique solution to the offset of the thunk.  I'd rather have riscv64 be consistent with all the other platforms and adjust the offset to the thunk in thunkEntry, not in specific code in Smalltalk.  Can you live with that?  The justification is that when we move the thunk initialization code down into a primitive the Smalltalk code won't change as much (obviously instance creation will be changed).  The way things are now riscv64 would be a special case.  But what I want is that all platforms simply invoke a primitive to get a new thunk and they use the address answered by the primitive without alteration.  Then it is up to the platform whether the adjustment is done in the primitive or in thunkProcess.  But it is *not* done  at the Smalltalk level.

KenDickey:
On 2022-08-03 15:01, Eliot Miranda wrote:

> @ELIOTMIRANDA commented on this pull request.
..
> Ken, I feel uneasy with the unique solution to the offset of the thunk. 
> I'd rather have riscv64 be consistent with all the other platforms and 
> adjust the offset to the thunk in thunkEntry, not in specific code in 
> Smalltalk. Can you live with that?

Sure.  Just remove the overwritten method.   If you look at the latest, 
the offset is zero.

-KenD

eliotmiranda:
On Wed, Aug 3, 2022 at 5:23 PM Ken Dickey ***@***.***> wrote:

> On 2022-08-03 15:01, Eliot Miranda wrote:
>
> > @ELIOTMIRANDA commented on this pull request.
> ..
> > Ken, I feel uneasy with the unique solution to the offset of the thunk.
> > I'd rather have riscv64 be consistent with all the other platforms and
> > adjust the offset to the thunk in thunkEntry, not in specific code in
> > Smalltalk. Can you live with that?
>
> Sure. Just remove the overwritten method. If you look at the latest,
> the offset is zero.
>

OK.  Fix thunkEntry to put back the offset and then I can merge the pull
request.

_,,,^..^,,,_
best, Eliot

KenDickey:
On 2022-08-04 12:09, Eliot Miranda wrote:

> On Wed, Aug 3, 2022 at 5:23 PM Ken Dickey ***@***.***> wrote:
..
>> Sure. Just remove the overwritten method. If you look at the latest,
>> the offset is zero.

> OK. Fix thunkEntry to put back the offset and then I can merge the pull
> request.

riscv64abicc.c as pull request should have
...
   vmcc.thunkp = thunkp;
...
which works fine without override of Callback>>addToThunkTable.  No 
offset required in either place.

So color me confused.

I can re-push the Alien-Core-Kend without the #addToThunkTable override 
if you like, but the current pull requiest should work with 
Alien-Core-Kend.124 as is.

How about accepting the pull and the Alien-Code, let me re-retest, and 
we will fix any breakage from there.

Epsilon close,
-KenD

eliotmiranda (approved):

KenDickey:
Tested riscv64 partial register dump via SIGSEGV -- looks OS (sqUnixMain.c)

  kill -s 11 <squeakPID>

PR #658, opened 2022-11-14
Fix gdi leak

nicolas-cellier-aka-nice:
On some win32 configurations and/or with some specific image,
we observe a GDI leak : numbers of GDI objects rapidly grow up to 10,000 (about 50 new objects/second).
when the limit is reached, the image hangs and we have to kill it...

With those fixes, the number of GDI objects remains stable at 20.

### DIFF
--- a/platforms/win32/vm/sqWin32Window.c
+++ b/platforms/win32/vm/sqWin32Window.c
@@ -1794,7 +1794,12 @@ static thisGetDpiForMonitor_t thisGetDpiForMonitor = NULL;
  */
 static double getDpiSystem(void)
 {
-  return (double) GetDeviceCaps(GetWindowDC(stWindow), LOGPIXELSY);
+  double dpi;
+  HDC dc = GetWindowDC(stWindow);
+  if (!dc) return 0.0; /* fail */
+  dpi = (double) GetDeviceCaps(dc, LOGPIXELSY);
+  ReleaseDC(stWindow,dc);
+  return dpi;
 }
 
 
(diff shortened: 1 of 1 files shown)


platforms/win32/vm/sqWin32Window.c
```diff
@@ -1793,7 +1794,12 @@ static thisGetDpiForMonitor_t thisGetDpiForMonitor = NULL;
  */
 static double getDpiSystem(void)
 {
-  return (double) GetDeviceCaps(GetWindowDC(stWindow), LOGPIXELSY);
+  double dpi;
+  HDC dc = GetWindowDC(stWindow);
+  if (!dc) return 0.0; /* fail */
```
marceltaeumel, line 1799: Note that `Float nan` is the contract for unknown dpi scale.

marceltaeumel: Nevermind. I mixed up dpi with scale factor here. Sorry for the noise.

marceltaeumel:
While I could not reproduce this issue on my system, I just verified that these changes do not change it for the worse. 😄 Works! Thank you.

nicolas-cellier-aka-nice:
I think that we can trigger the Bug by forcing usage of a CursorWithAlpha, maybe thru:

    Cursor useBiggerCursors: true.

or

    2000 timesRepeat:
        [CursorWithAlpha biggerNormal show.
        Project current world doOneCycle.
        20 milliSeconds wait].

(BTW, I hate milliSeconds and would prefer milliseconds as it is a single word)

marceltaeumel:
>I think that we can trigger the Bug by forcing usage of a CursorWithAlpha, maybe thru:

Nope. Does not show on my machine. Stable at 20 GDI objects.

eliotmiranda:
Thanks Nicolas!!

_,,,^..^,,,_ (phone)

PR #660, opened 2022-11-30
configure.ac: Fix tzet typo for tzset

fweimer-rh:
Found with an instrumented compiler for:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


krono:
Marvelous!

Issue #665, opened 2023-08-11
sbrk() sqUnixSpurMemory patch


cstes:
Hi,

I have sent in the past email to Eliot Miranda about an issue which I discovered in August 2021 (two years ago), and I thought to raise an issue here.

It is a Linux/UNIX patch issue with the behavior of the mmap() system call or library call as used by the OpenSmalltalk COG vm.

When way back in 2021 some changes were made for Linux ARM8 to the file saUnixSpurMemory.c I reported that this was breaking the Solaris 11.3 build.   Curiously the Solaris 11.4 build was still working fine.

I think (I don't know) that perhaps the behavior of mmap() on Solaris 11.4 is slightly diferent than on 11.3, to match (perhaps) the Linux kernel behavior of mmap().

An idea that I have is to ask some UNIX guru's whether there exist a autoconf-archive (configure script) test or script to test the behavior of mmap() for portability.

However here I will just report what the issue is :   since august 2021 I am building succesfully on Solaris with the following patch
```

--- opensmalltalk-vm-sun-v5.0.34/platforms/unix/vm/sqUnixSpurMemory.c	Fri Aug 13 19:54:42 2021
+++ p0/opensmalltalk-vm-sun-v5.0.34/platforms/unix/vm/sqUnixSpurMemory.c	Sun Aug 15 14:11:58 2021
@@ -277,16 +277,21 @@
 {
 	void *hint = sbrk(0); // a hint of the lowest possible address for mmap
 	void *result;
+	char *address;
+	unsigned long alignment;
 
 	pageSize = getpagesize();
 	pageMask = ~(pageSize - 1);
 
+	alignment = max(pageSize,1024*1024);
+	address = (char *)(((usqInt)hint + alignment - 1) & ~(alignment - 1));
+
 #if !defined(MAP_JIT)
 # define MAP_JIT 0
 #endif
 
 	*desiredSize = roundUpToPage(*desiredSize);
-	result =   mmap(hint, *desiredSize,
+	result =   mmap(address, *desiredSize,
 #if DUAL_MAPPED_CODE_ZONE
 					PROT_READ | PROT_EXEC,
 #else
```

I really need Eliot's help on this.

Eliot : what I did was try to understand your code and you seem to make some alignment calculations on the desired memory location that you request through mmap().

I believe that the traditional UNIX mmap() or the Solaris 10 and Solaris 11.3 mmap() is reacting slightly differently to the requested memory "hint".

[...]

PR #672, opened 2024-01-06
align return value of sbrk()

dcstes:
Possible fix for issue mmap() sqUnixSpurMemory patch #665

David Stes

### DIFF
--- a/platforms/unix/vm/sqUnixSpurMemory.c
+++ b/platforms/unix/vm/sqUnixSpurMemory.c
@@ -277,16 +277,21 @@ allocateJITMemory(usqInt *desiredSize)
 {
 	void *hint = sbrk(0); // a hint of the lowest possible address for mmap
 	void *result;
+	char *address;
+	unsigned long alignment;
 
 	pageSize = getpagesize();
 	pageMask = ~(pageSize - 1);
+
+	alignment = max(pageSize,1024*1024);
+	address = (char *)(((usqInt)hint + alignment - 1) & ~(alignment - 1));
 
 #if !defined(MAP_JIT)
 # define MAP_JIT 0
 #endif
 
 	*desiredSize = roundUpToPage(*desiredSize);
-	result =   mmap(hint, *desiredSize,
+	result =   mmap(address, *desiredSize,
 #if DUAL_MAPPED_CODE_ZONE
 					PROT_READ | PROT_EXEC,
 #else


eliotmiranda:
Hi David, can you try a simpler version of the patch? That is just add

    hint = roundUpToPage(hint);

before hint is used. So that could go on the line after hint = sbrk(0);

cstes:
I believe that I tried that simpler patch and that worked, meaning the loop did not end with an intialAlloc failed message, but my PR (pull reauest) also works and it also means that the loop does not end with an initialAlloc failed.   I have to look at the code in sqAllocateMemorySegmentOfSizeAboveAllocatedSizeInto()   and at the code for using sbrk() in the case of the stack interpteter.

cstes:
There is a call to sbrk() in sqAllocateMemory() but there is another call to sbrk() in allocateJITMemory().  So there are multiple calls to sbrk() in platforms/unix/vm/sqUnixSpurMemory.c.   The pull request #672 adds an alignment calculation to the return value of sbrk() in allocateJITMemory() so to mirror the calculation in sqAllocateMemory().   The OpenGroup manpage on sbrk() says that it is unspecified whether the pointer returned by sbrk() is aligned suitably for any purpose  (see https://pubs.opengroup.org/onlinepubs/7908799/xsh/brk.html).

eliotmiranda:
We should abstract it into a function which names its purpose, eg lowestPageAlignedAddressForSegmentAllocation.I’ll do this today._,,,^..^,,,_ (phone)On Jan 27, 2024, at 6:10 AM, David Stes ***@***.***> wrote:﻿
There is a call to sbrk() in sqAllocateMemory() but there is another call to sbrk() in allocateJITMemory().  So there are multiple calls to sbrk() in platforms/unix/vm/sqUnixSpurMemory.c.   The pull request #672 adds an alignment calculation to the return value of sbrk() in allocateJITMemory() so to mirror the calculation in sqAllocateMemory().   The OpenGroup manpage on sbrk() says that it is unspecified whether the pointer returned by sbrk() is aligned suitably for any purpose  (see https://pubs.opengroup.org/onlinepubs/7908799/xsh/brk.html).

eliotmiranda:
On Jan 28, 2024, at 11:01 AM, Eliot Miranda ***@***.***> wrote:﻿We should abstract it into a function which names its purpose, eg lowestPageAlignedAddressForSegmentAllocation.I’ll do this today.I’ve tried this and something strange happened on my Mac; changing the hint appears to have changed the address of the initial allocation to something much higher, which causes the existing code to fail to allocate (the validation loop in the main segment allocator).  So this is taking longer than anticipated. _,,,^..^,,,_ (phone)On Jan 27, 2024, at 6:10 AM, David Stes ***@***.***> wrote:﻿
There is a call to sbrk() in sqAllocateMemory() but there is another call to sbrk() in allocateJITMemory().  So there are multiple calls to sbrk() in platforms/unix/vm/sqUnixSpurMemory.c.   The pull request #672 adds an alignment calculation to the return value of sbrk() in allocateJITMemory() so to mirror the calculation in sqAllocateMemory().   The OpenGroup manpage on sbrk() says that it is unspecified whether the pointer returned by sbrk() is aligned suitably for any purpose  (see https://pubs.opengroup.org/onlinepubs/7908799/xsh/brk.html).

cstes:
Hi Eliot,
Thanks for looking at this issue and pull request.   Also the fact that I post it on github does not mean you have to immediately work on it.   But I wanted to remind you on this issue so that hopefully you find a good solution.   I appreciate your work on OpenSmalltalk and your knowledge on the Smalltalk environment will be of great help here.   Working on such a large and complex piece of software and looking at the feedback can be a lot of work and I only want to give some feedback, not increase your (pesumably high) workload ...  Based on the feedback from David Lewis I understand thatthe pull request modification works on Linux.   Because the specificatoin of sbrk() in the SUS (Single UNIX Specification) is probably applicable to all sbrk() implementations on all UNIX based operating systems, I think it is likely to work on all UNIX based platforms.
Regards

dcstes:
solved by new code lowestPageAligendAddressForMMap() macro in Spur memory management for Unix

PR #673, opened 2024-01-06
use --whole-archive option instead of -whole-archive for solaris ld

dcstes:
I believe that the GNU link editor supports both single hyphen and double hyphen options -whole-archive and --whole-archive

However the SUN (Solaris and SunOS and Illumos) link editor uses different names like "-z weakextract"

Three is a GNU compatibility layer but the SUN link editor insists for GNU compatibility on the double hyphen format -whole-archive and --no-whole-archive

David Stes

### DIFF
--- a/platforms/unix/config/Makefile.in
+++ b/platforms/unix/config/Makefile.in
@@ -45,7 +45,7 @@ all : $(squeak) plugins squeak.1 $(npsqueak)
 # VM
 
 $(squeak) : config.sh $(SQLIBS) version.o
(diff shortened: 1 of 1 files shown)


dcstes:
Any news on this ? Preferably this PR would be merged by the author of the Makefile modification that uses "whole-archive".

dtlewis290:
Tobias added the whole-archive flags a few years ago. This PR seems reasonable and correct so I see no reason not to merge it. If there is still some platform that uses 'whole-archive' we can easily revert the change.

dtlewis290:
Correcting a possibly confusing typo, I meant to say "If there is still some platform that uses '-whole-archive' we can easily revert the change"

PR #686, opened 2024-06-20
Check bind and listen functions for errors

jraiford1:
When listening on a socket, if the local port is already in use it should fail, however the vm currently assumes that the listen() call is successful. This patch implements a similar check as the connect call.  Note that the Windows implementation does not have this issue.


eliotmiranda (approved):

eliotmiranda:
Looks good. Thanks Jon & welcome!

PR #689, opened 2024-07-09
remove check for socket file presence in sqResolverGetAddressInfoHost…

nafiz1001:
…SizeServiceSizeFlagsFamilyTypeProtocol


nafiz1001:
I just realized I could have just replaced the `&&` with an `||` instead of removing the file check...

PR #697, opened 2024-12-05
Makes dropRequestFileName percent-decode on Unix.

### LINKED ISSUE #666: DropPlugin should percent-decode the file URI on Unix
On Unix, when dropping a file that has spaces or other special characters in its path, `primitiveDropRequestFileName` returns a [percent-encoded](https://en.wikipedia.org/wiki/Percent-encoding) version of the path and `primitiveDropRequestFileHandle` fails.

I think the problem is that [`dropRequestFileName`](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c#L72-L101) [skips the `file:///` prefix of the URI](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c#L94-L95), but does not percent-decode it. Interestingly, [`uri2string`](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/platforms/unix/vm-display-X11/sqUnixXdnd.c#L206-L234), which [isn't currently used](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/platforms/unix/vm-display-X11/sqUnixXdnd.c#L204-L205), [does percent-decoding](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/platforms/unix/vm-display-X11/sqUnixXdnd.c#L219-L223).

`dropRequestFileName` [is used by `primitiveDropRequestFileName`](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/src/plugins/DropPlugin/DropPlugin.c#L132) and also by [`dropRequestFileHandle`](https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/330d6779ad2ecbce1f07131f53d75cd168165f0b/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c#L117), which is why `primitiveDropRequestFileHandle` fails.

Note: I did not check whether this problem exists on other platforms.

valnssh:
Resolves https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/666, which describes a problem on Unix systems where dropping a file with spaces or special characters in its path results in a percent-encoded version of the path being returned by `primitiveDropRequestFileName`.

The issue occurs because `dropRequestFileName` skips the `file:///` prefix of the URI but does not perform percent-decoding on the path.

### DIFF
--- a/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c
+++ b/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c
@@ -41,6 +41,7 @@
 #include "sqVirtualMachine.h"
 #include "FilePlugin.h"
 #include "DropPlugin.h"
+#include <ctype.h>
 
 extern struct VirtualMachine  *interpreterProxy;
 extern int						uxDropFileCount;
--- a/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c
+++ b/platforms/unix/plugins/DropPlugin/sqUnixDragDrop.c
@@ -66,6 +66,18 @@ fileValueOf(sqInt anSQFileRecord)
 sqInt dropInit(void)     { return 1; }
 sqInt dropShutdown(void) { return 1; }
 
+static int
+hexValue(const int c)
+{
+  if (c <  '0') return 0;
+  if (c <= '9') return c - '0';
+  if (c <  'A') return 0;
+  if (c <= 'F') return c - 'A' + 10;
+  if (c <  'a') return 0;
+  if (c <= 'f') return c - 'a' + 10;
+  return 0;
+}
+
 /* We now set USE_FILE_URIs to 1 in platforms/unix//vm-display-X11/sqUnixXdnd.c
  * hence dropRequestFileName skips the URI prefix and dropRequestURI includes it
  */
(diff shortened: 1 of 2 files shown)


eliotmiranda (commented):
This looks fine. The only thing I would quibble with is the use of strlen(path + i + 3). Since we know the length up front we could use e.g.

```
for (int i = 0, l = strlen(path); path[i]; ++i, --l) {
		if ((path[i] == '%') && isxdigit(path[i + 1]) && isxdigit(path[i + 2])) {
			path[i] = hexValue(path[i + 1]) << 4 | hexValue(path[i + 2]);
			l -= 2;
			memmove(path + i + 1, path + i + 3, l);
		}
	}
```
Not sure if my suggestion is exactly correct but you get the idea.  If you like the suggestion then submit a revised commit, otherwise LMK and I'll approve as is.

valnssh:
Hi @eliotmiranda!

You're absolutely right. I prefer your approach, not only for performance reasons but also because I find it more declarative and easier to understand. I appreciate the suggestion, and I'll be pushing a commit with these changes right now.

Thanks!

eliotmiranda (approved):

PR #700, opened 2025-01-25
configure script option to build without vm-sound-Sun plugin

dcstes:
On Solaris, "configure" checks the header file sys/audioio.h and if it finds it enables support for vm-sound-Sun.

This patch introduces an option for configure (--without-vm-sound-Sun) so that support for vm-sound-Sun can be disabled even when the header file is present.

This change affects all unix platforms as "configure" is used on all unix builds.

I tested the patch by running "gmake configure" in platforms/unix/config and by building in the building/sunos cog-spur and stack-spur building directories (by running the "mvm" script).

### DIFF
--- a/platforms/unix/vm-sound-Sun/acinclude.m4
+++ b/platforms/unix/vm-sound-Sun/acinclude.m4
@@ -1,5 +1,11 @@
 # -*- sh -*-
 
+AC_ARG_WITH(vm-sound-Sun,
+  AS_HELP_STRING([--without-vm-sound-Sun],[disable Sun vm sound support (default=enabled)]),
+  [with_vm_sound_Sun="$withval"],
+  [with_vm_sound_Sun="yes"])
+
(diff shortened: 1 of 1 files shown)


dtlewis290:
Affects only Sun platform, and I confirmed no issues for building on Linux. Follow up is required to rebuild the configure script (cd platforms/unix/config; make). I had no issue doing this, but it would be best if Tobias could commit the update for configure in order to minimize noise in the commit log.

krono:
needs `make` in unix config dir to re-gen `configure`, tho

dtlewis290:
Yes, it needs  make in unix config dir. @krono do you have time to do this? I am able to do it but my config tools are at different versions that would create noise in the diffs (and maybe some actual problems, I do not know). So it would be best if this can be done with the same tools that were used for previous updates. Thanks.

krono:
sorry im late.
I ran the re-gen

PR #722, opened 2025-05-20
vm-sound-Sun: replace AC_TRY_COMPILE by AC_COMPILE_IFELSE

### LINKED ISSUE #723: configure autoconf 2.72 : AC_TRY_COMPILE and AC_COMPILE_IFELSE
autoconf 2.72 claims that AC_TRY_COMPILE is obsolete.
it may be possible to replace this macro by AC_COMPILE_IFELSE in some cases, or by other macros (AC_CHECK_HEADERS or so)

https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html#index-AC_005fTRY_005fCOMPILE

In section: 18.4 Obsolete Macros
Several macros are obsoleted in Autoconf, for various reasons (typically they failed to quote properly, couldn’t be extended for more recent issues, etc.). They are still supported, but deprecated: their use should be avoided.

apparently AC_TRY_COMPILE is still accepted and it seems autoconf 2.72 only prints a warning

dcstes:
this pull request/commit only changes the vm-sound-Sun plugin

autoconf 2.72 claims that AC_TRY_COMPILE is obsolete.

https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html#index-AC_005fTRY_005fCOMPILE

In section: _18.4 Obsolete Macros
Several macros are obsoleted in Autoconf, for various reasons (typically they failed to quote properly, couldn’t be extended for more recent issues, etc.). They are still supported, but deprecated: their use should be avoided._

apparently AC_TRY_COMPILE is still accepted and it seems autoconf 2.72 only prints a warning

however, it may be a good idea to move to AC_COMPILE_IFELSE

this commit replaces the old AC_TRY_COMPILE which I think comes from old code from original Squeak developers by a more recent variation which is intented to do the same thing

### DIFF
--- a/platforms/unix/vm-sound-Sun/acinclude.m4
+++ b/platforms/unix/vm-sound-Sun/acinclude.m4
@@ -1,4 +1,5 @@
 # -*- sh -*-
+# test whether the macro AUDIO_SUNVTS is defined in <sys/audioio.h> or <sun/audioio.h>
 
 AC_ARG_WITH(vm-sound-Sun,
   AS_HELP_STRING([--without-vm-sound-Sun],[disable Sun vm sound support (default=enabled)]),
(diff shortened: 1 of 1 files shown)


dcstes:
Perhaps it is better to use AC_PREPROC_IFELSE

Please do not merge this PR , although the replacement of AC_TRY_COMPILE by AC_COMPILE_IFELSE is a possibility, there is perhaps a better way.

PR #725, opened 2025-05-22
Add build files for linux64x64/squeak.sista.spur

LinqLover:
Not sure if I have extrapolated all flags correctly from squeak.cog.spur and linux32x86, but they worked for me and @MariusDoe.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 platforms/unix/config/autom4te.cache/
 
 # Don't track build directory contents (with exceptions)
+# NB: This logic is flawed (but compact). When adding new building scripts, commit them via `git add -f` or change the .gitignore temporarily.
 /building/**/**/*
 # except do track these files
 !/building/**/common*/*
--- a/building/linux64x64/squeak.sista.spur/build.assert/mvm
+++ b/building/linux64x64/squeak.sista.spur/build.assert/mvm
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+# assert Sista Spur VM with VM profiler and threaded heartbeat
+INSTALLDIR=assert/sqsisspur64linuxht
+OPT="-g3 -O1 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -DDEBUGVM=0"
+
+# Prefer clang over gcc, but use gcc if clang isn't available...
+CC=gcc
+if [ -x /usr/bin/clang ]; then CC=clang; fi
+
+if [ $# -ge 1 ]; then
+	INSTALLDIR="$1"; shift
+fi
+if ../../../../scripts/checkSCCSversion ; then exit 1; fi
+echo -n "clean? "
+read a
+case $a in
+n|no|N|NO)	echo "ok but this isn't safe!!";;
+*)			test -f Makefile && make reallyclean
+esac
+../../../../scripts/copylinuxpluginspecfiles
+test -f config.h || ../../../../platforms/unix/config/configure \
+		--with-vmversion=5.0 \
+		--with-src=src/spur64.sista \
+		--with-scriptname=spur64 \
+	TARGET_ARCH="-m64" \
+	CC=$CC \
+	CFLAGS="$OPT -msse2 -DCOGMTVM=0"
+rm -f vm/sqUnixMain.o # nuke version info
+rm -rf ../../../../products/$INSTALLDIR
+# prefer make install prefix=`readlink -f \`pwd\`/../../../../products/$INSTALLDIR`
+# but older linux readlinks lack the -f flag
+make install-squeak install-plugins prefix=`(cd ../../../../;pwd)`/products/$INSTALLDIR 2>&1 | tee LOG ; test ${PIPESTATUS[0]} -eq 0
--- a/building/linux64x64/squeak.sista.spur/build.debug/mvm
+++ b/building/linux64x64/squeak.sista.spur/build.debug/mvm
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+# debug Sista Spur VM with VM profiler and threaded heartbeat
+INSTALLDIR=debug/sqsisspur64linuxht
+OPT="-g3 -O0 -DDEBUGVM=1"
+
+# Prefer clang over gcc, but use gcc if clang isn't available...
+CC=gcc
+if [ -x /usr/bin/clang ]; then CC=clang; fi
+
+if [ $# -ge 1 ]; then
+	INSTALLDIR="$1"; shift
+fi
+if ../../../../scripts/checkSCCSversion ; then exit 1; fi
+echo -n "clean? "
+read a
+case $a in
+n|no|N|NO)	echo "ok but this isn't safe!!";;
+*)			test -f Makefile && make reallyclean
+esac
+../../../../scripts/copylinuxpluginspecfiles
+test -f config.h || ../../../../platforms/unix/config/configure \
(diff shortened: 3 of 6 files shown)


codefrau:
The sista builds are still disabled:

https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/1f9cfccaed7b7f660a08618c3b9e86dec01117ea/.github/workflows/linux.yml#L65

Was there another issue, or was it just the missing build files?

(BTW, what is squeak.sista.spur? squeak.cog.spur appears to run Sista code just fine – like, we still don't have squeak.sista.spur for Mac)

LinqLover:
> The sista builds are still disabled:
> 
> https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/1f9cfccaed7b7f660a08618c3b9e86dec01117ea/.github/workflows/linux.yml#L65
> 
> Was there another issue, or was it just the missing build files?

Thanks, will try!

> 
> (BTW, what is squeak.sista.spur? squeak.cog.spur appears to run Sista code just fine – like, we still don't have squeak.sista.spur for Mac)

squeak.cog.spur does interpret the Sista bytecode set, but that's just the first part of the SistaVM. @MariusDoe and I are currently taking the first steps of reviving Scorch (an adaptive bytecode optimizer for Sista by @clementbera), porting it from Pharo to Squeak, and building tooling for debugging and extending it. Concretely, the Sista VM collects additional statistics (`primitiveSistaMethodPICAndCounterData`) and notifies the running image when data are available via `Context>>conditionalBranchCounterTrippedOn:`, where the adaptive optimizer can use them to create optimized CompiledMethods with inlined branches and primitives. For more details, please ask. :-)

Best, 
Christoph

LinqLover:
PS:

> we still don't have squeak.sista.spur for Mac

Sorry, but unless Apple becomes more virtualization-friendly, Apple users will have to accept that at least I won't be putting any voluntary effort into supporting the code I write for their platforms.

krono:
what does that have to do with virtualization?

LinqLover:
> what does that have to do with virtualization?

Without proper virtualization support, non-Apple users cannot easily test and debug their software on macOS. Virtualizing Linux from a host Windows or macOS usually works well. Virtualizing Windows from a host Linux or macOS still works kind of well, even though their new TPM requirement makes it more difficult. Virtualizing macOS from a host Linux/Windows only works via unofficial Hackintosh, is said to be unstable, and is also illegal. So a) trying to support any of my code for macOS would be extra effort for me as a Linux user (even more than supporting Windows) and bear legal risks and b) it goes against my open-source ethos to support such a closed ecosystem run by a company that abuses their quasi-monopoly position to lock-in developers and stifle open development. :-)

krono:
> it goes against my open-source ethos to support such a closed ecosystem run by a company that abuses their quasi-monopoly position to lock-in developers and stifle open development. :-)

Which one? MS or Apple ;)

anyway, I get now which part you mean.

LinqLover:
> Which one? MS or Apple ;)

In my impression Apple is even worse than Microsoft. Microsoft at least has a significant open source portfolio for developers, even though they surely don't do it for altruism.

krono:
[moving discussion to elsewhere]

krono:
@codefrau 
> (BTW, what is squeak.sista.spur? squeak.cog.spur appears to run Sista code just fine –

There's a cog source:
 - https://github.com/OpenSmalltalk/opensmalltalk-vm/tree/Cog/src/spur64.cog
and a sista one:
- https://github.com/OpenSmalltalk/opensmalltalk-vm/tree/Cog/src/spur64.sista

One uses the `StackToRegisterMappingCogit` the other the `SistaCogit`; the latter defines 3 more primitves etc.pp.
not counting the CogMT variant, the sista src is some 17k more lines of code.

So the cog one can run sista images, but the sista one has the jit for it.

> like, we still don't have squeak.sista.spur for Mac)

 I do see build paths for both [intel](https://github.com/OpenSmalltalk/opensmalltalk-vm/tree/Cog/building/macos64x64/squeak.sista.spur) and [silicon](https://github.com/OpenSmalltalk/opensmalltalk-vm/tree/Cog/building/macos64ARMv8/squeak.sista.spur). so do you mean the ci builds?

eliotmiranda:
> (BTW, what is squeak.sista.spur? squeak.cog.spur appears to run Sista code just fine – like, we still don't have squeak.sista.spur for Mac)

As Christof (LinqLover) said squeak.sista.spur would be a build directory for the full Sista (Speculative Inlining Smalltalk Architecture) scheme. This is Clément’s thesis. See the publications page on squeak.org.

the Sista bytecode set is two things, a better designed set of bytecodes for normal Smalltalk execution (which is what we’re using now), plus a number of additional bytecodes for generating fast unsafe code, principally a set of unsafe inline primitives.

The Bytecode compiler and the opensmalltalk-vm JIT (StackToRegisterMappingCogit) generate/JIT the normal Smalltalk part of the SistaV1 bytecode set (specified by the class comment of EncoderForSistaV1, worth a read).

The SistaCogit extends StackToRegisterMappingCogit with support for the speculative inlining/adaptive optimization part of the Sista bytecodes, in particular planting counting code at conditional branches and generating code for the unsafe inline primitives. When a counter trips a message is sent to the image, intended to invoke the Scirch bytecode-to-bytecode optimiser. Scorch uses the introspection primitives implemented by SistaCogit to fetch type information from the inline caches in the machine code and uses this to generate replacement compiled methods that
- inline compiled methods
- use the unsafe inline primitives to access state/invoke functionality without the overheads in the safe versions 

Christof and Marius, as part of their masters theses, are reviving Clément’s Scorch optimiser. The main focus right now is in making Scorch unaware of its location. Scorch is designed to sit alongside the rest of the image, a sibling of the bytecode compiler. But it is too unreliable at the moment for this to be safe and for it to be a workable development context. But by implementing a proxy framework Marius and Christoph are able to put Scorch alongside the simulator.

The simulator is configured to use the SistaCogit, counter trips happen in the simulated image, but are delivered to a Scorch in the host image, a sibling of the simulator. It accesses the simulated image, unaware of that fact because the proxies behave just like contexts ;et al) in the host image. So bugs in code generated by Scorch affect only the simulated image, not the host.

Christoph and Marius in only a few weeks have revived the project, built the proxy framework, and are now using it to develop Scorch. Clément’s has joined us in a meeting and has agreed to provide mentorship.  So we have a great chance at realizing Sista, which should produce a 3x to 4x speed up in its first production incarnation.

To reach the first incarnation we have to
- fix code generation bugs in Scorch
- implement an atomic deoptimizer (a deoptimizer that can be expressed via Slang and put in the vm) to avoid potential recursive de optimization errors [this may not be necessary, but is a safe approach; deoptimization is relatively straightforward, guided by suitable metadata generated by Scorch]
- extend SistaCogit with improved register allocation. StackToRegisterMappingCogit allocates registers only for the receiver and arguments of 9, 1, & 2 argument sends. The improved JIT will also allocate registers to temporaries. The essential point is that Scorch’s inlining generates much larger basic blocks so allocating registers to temporaries is profitable.

HTH

LinqLover:
Thanks for the excellent summary, Eliot! I hope we will be able to share more detailed updates on the list soon. :-)

> Christof and Marius, as part of their masters theses, are reviving Clément’s Scorch optimiser.

(Just to be precise, I already have written my master's thesis about a different topic - semantic exploratory programming, https://github.com/LinqLover/semexp-thesis -, but the Sista project might be relevant for my PhD thesis :D)

Best,
Christoph

PR #728, opened 2025-06-24
[ci] add squeak.sista.spur build for linux64x64

LinqLover:


krono:
@eliotmiranda any opposition to add this to the ci? is it meaningful?

eliotmiranda:
No. But I don’t think sista is worth building yet. I hops to productise it over the next two or three years. Christoph and Marius pushed things forward so that we can develop it in the simulator. But the development still has to be finished.

PR #731, opened 2025-07-09
Updated for OpenBSD 7.7

KenDickey:
Built & tested on "build", "build.debug" and "build.assert" for both stack and cog VMs on 
- Raspberry Pi 5 Raspian Linux [Aarch64]
- Raspberry Pi 4 OpenBSD 7.7 [Aarch64]
- Optiplex 350 OpenBSD 7.7 [amd64]

Does NOT build Camera plugin. 
[Needs conditional for !__OpenBSD__ ; note: `uname` -> "OpenBSD"]

Passes base image and FFI unit tests on Cuis current (ver7.5 rev 7279).
OpenBSD runs current Squeak [6.0-22148-64bit] but fails some unit tests.  SSL not working.
Unable to test on RasPi/Raspian-Linux due to Window Continual Resize bug.
Squeak unit tests see symptom on console:
  sqAllocateMemorySegmentOfSizeAboveAllocatedSizeInto mmap: Cannot allocate memory
No such symptom with Cuis.

Note: [Forgot that stack vm does not use dual mapped memory; stack mvm's update spurious]


KenDickey:
SqueakSSL plugin update for OpenBSD.
Now Squeak image updates code.
Can build spur64 VMMaker [both Linux & OpenBSD]
WebClient unit tests pass [amd64+OBSD, arm64+Linux, arm64+OBSD].
Get same 10 unit test errors for Linux as OpenBSD for SqueakSSL.
I think we should merge this so when next released I can create an OpenBSD port.

eliotmiranda:
> SqueakSSL plugin update for OpenBSD. Now Squeak image updates code. Can build spur64 VMMaker [both Linux & OpenBSD] WebClient unit tests pass [amd64+OBSD, arm64+Linux, arm64+OBSD]. Get same 10 unit test errors for Linux as OpenBSD for SqueakSSL. I think we should merge this so when next released I can create an OpenBSD port.

+1

cstes:
My comment does not relate directly to OpenBSD, but between 2021 and 2023 there was on Solaris 11.3 (SVR4 unix) an issue with sqAllocateMemorySegmentOfSizeAboveAllocatedSizeInto mmap: Cannot allocate memory.   This related to sbrk() - see issue #672 and #665.  This was subsequently fixed (by Eliot) so currently (recently) I do not experience that error any more.    However I read in this issue that there is a difference between Cog and Stack VM by using "dual mapped memory" and the mvm script seems to be modified to set a variable for DUALMAP.  Should this variable DUALMAP be used please ?

eliotmiranda:
> … there is a difference between Cog and Stack VM by using "dual mapped memory" and the mvm script seems to be modified to set a variable for DUALMAP. Should this variable DUALMAP be used please ?

it depends :-) Dual mapping is one way of editing code in JIT code zones on OSs that disallow allocating read+write+execute memory. The idea is to maintain a read+execute zone in which the executable code is kept, but map this zone elsewhere in the address space as read+write, and write code to this mapping. Another way is to change permissions from read+execute to read*write around any code editing operation. The former means is more efficient but may be defeated by the OS (as is the case on macOS). So whether the scheme is needed or not depends on the specific OS.

KenDickey:
OpenBSD implements X^W where memory has only 1 of eXecute xor Write permission.
To allow Cog's trick of dual mapping, OpenBSD requires the opensmalltalk-vm be built rooted in /usr/local.  {Same for other JITs}
OpenBSD also requires the linker flag "wxneeded".  I believe a couple of other flags & include files differ.

The Stack VM does not need or use dual mapping.

KenDickey:
Conflicts in mvm files due to 
- [1] def/flag changes -- should be accepted [IMHO]
- [2] CC=clang
Perhaps better with the linux64x64 solution:
vvv===vvv===vvv
# Prefer clang over gcc, but use gcc if clang isn't available...
CC=gcc
if [ -x /usr/bin/clang ]; then CC=clang; fi
^^^===^^^===^^^

KenDickey:
Built with clang (rather than gcc) works fine on RasPi4+OpenBSD but RasPi5+RaspianLinux fails to build blitbit plugin with my OpenBSD fork.  Perhaps this works with current due to changes in bitblit plugin.  ?? 
FYI

KenDickey:
RaspianLinux+RasPi5 : Aarch64 Linux : removed "-enable-fast-bitblt" build flag 
- Cog: "build.assert" & "build.debug" work using clang; "build" fails
- Stack: "build" works with clang
- 
Note: "build" works with clang on OpenBSD+RasPi4 [also Aarch64]

Details atached:

[clangCogArm64-bug.txt](https://github.com/user-attachments/files/21633944/clangCogArm64-bug.txt)

KenDickey:
I should clarify.
squeak.cog.spur/build/mvm for Aarch64:
- VM works when using gcc; builds but segfaults with clang [Linux]
- VM works with either gcc or clang on OpenBSD
- again, VM on Linux works with either clang or gcc for "build.debug" & "build.assert"

KenDickey:
On 2025-08-06 16:27, OpenSmalltalk-Bot wrote:

> OPENSMALLTALK-BOT left a comment (OpenSmalltalk/opensmalltalk-vm#731) I 
> most strongly urge anyone feeling brave to make sure that the 
> enable-fast-bitblt can be used because it includes some very dramatic 
> performance improvements for important use cases. There should be a 
> pretty serious pile of test code in the repository that Ben Avison used 
> when developing this and that should help.
> 

Thanks, Tim.

My goal in removing the fast-bitblt flag is to get a viable build with 
clang to avoid a problem noted by Marcel Taeumel

https://github.com/KenDickey/opensmalltalk-vm/commit/e0c5e12ed8889d4195e47773c0210680d579fccd

The problem with this is that clang chokes on the fast bitblit code.

As you probably remember, I have hardly used C & tools for a couple of 
decades, so am not someone who should be investigating this.


The good news is that I got the VM to build properly with OpenBSD on 
both amd64 and arm64 and also got SSL code loading and WebClient unit 
tests to work with current Squeak image.  Also, VMMaker works on arm64 
Linux.

I really am agnostic on clang vs gcc as I seldom use either.  But I am 
willing to do shallow push-through where I can to get the VM running in 
more environments.  [OpenBSD port is Squeak 3.10 -- nice to have 
current].

I am slowly learning TLS 1.3 to work on getting SquealSSL unit tests 
working, but again this is waaaayy out of my comfort zone.  I work in 
Smalltalk for a reason!  ;^)

BTW: Where might Ben Avison's code be located?

Thanks for all,
-KenD

KenDickey:
With the latest, able to build on Aarch64 Linux [tested on Raspian & Manjaro] and OpenBSD.  

Upon Linux, uses gcc and --enable-fast-bitblt.

As builds & runs/tests cross platform, NOW would be a good time to integrate.  [IMHO]

PR #735, opened 2025-08-08
OBSD does not supply librt

KenDickey:
Not long ago, "librt" functions moved into "libc".  Most Linux distros supply empty "librt" library for back compatibility.
OpenBSD does not supply "librt" and this chokes the build on OBSD when "-lrt" asked of the linker.


KenDickey:
Ah.  Tested with:
- Manjaro & Raspian Linux [Aarch64]
- OpenBSD [amd64 & Aarch64]

PR #736, opened 2025-08-26
Better Framebuffer Notes for Alpine Linux

KenDickey:
I like Alpine Linux as a light/shim OS, but the setup takes some care.

Updated file:
-   platforms/unix/vm-display-fbdev/AlpineLinux-Notes.txt


krono:
thank you

PR #737, opened 2025-08-27
Framebuffer input-handling improvements

tonyg:
Hi Eliot, all,

Working recently with a stripped-down Alpine installation on an oldish Thinkpad (X240) with a touchpad, I found the framebuffer input handling support did not work. The core issue is that `/dev/input/event*` for the touchpad yields `ABS_*` events, not `REL_*` events.

This branch makes a few changes to (a) very slightly clean up and (b) improve `/dev/input` handling in the framebuffer driver.

 - Small sections of unused code removed.
 - Option for flipping option/command, alt/meta, super/alt, etc (because it was strangely configured out-of-the-box for me)
 - Crude (!) conversion of absolute multitouch input to relative, mouse-like input. Very simple tap detection, only deals with a single finger, not scroll wheel etc.

Potential improvements:
 - Timeout for deciding a tap isn't happening, rather than solely being based on moving outside a fixed radius
 - Configurable radius for motion hysteresis
 - Handling of two-, three- and four-finger taps as left/middle/right mouse button; at present, the code just gives a single left click, not even a click-and-hold, so modifier keys are required for other mouse buttons
 - Handling of Mac-style tap-and-hold for click-and-hold (should be fairly easy, requires interpretation of timestamps)
 - more?

Ultimately it's a bit of a shame this is done in C; I've had good results reading `/dev/input` *directly* from Squeak (for the Squeakphone), so I'd be interested in discussing ways we could lift more of Squeak's event handling into the image. Until then of course it'd be nice to improve the C support code we have. Thoughts welcome!

### DIFF
--- a/platforms/unix/vm-display-fbdev/sqUnixEvdevKeyMouse.c
+++ b/platforms/unix/vm-display-fbdev/sqUnixEvdevKeyMouse.c
@@ -111,6 +111,14 @@ struct ms
   char 		  *msName;
   int		   fd;
   struct libevdev *dev;
+
+  /* Emulation of relative mouse device using absolute device */
+  int touching;
+  int moved;
+  int has_x;
+  int has_y;
+  int curr_x;
+  int curr_y;
 };
 
 static struct ms mouseDev;
@@ -231,6 +239,19 @@ static void updateMouseButtons(struct input_event* evt) {
 	case BTN_LEFT:   buttonState |= LeftMouseButtonBit;  break;
 	case BTN_MIDDLE: buttonState |= MidMouseButtonBit;   break;
 	case BTN_RIGHT:  buttonState |= RightMouseButtonBit; break;
+
+        /* /\* When simulating a 3-button mouse via a multitouch touchpad: *\/ */
+        /* case BTN_TOOL_DOUBLETAP: buttonState |= LeftMouseButtonBit;  break; */
+        /* case BTN_TOOL_TRIPLETAP: buttonState |= RightMouseButtonBit; break; */
+        /* case BTN_TOOL_QUADTAP:   buttonState |= MidMouseButtonBit;   break; */
+
+        case BTN_TOUCH: {
+          mouseDev.touching = 1;
+          mouseDev.moved = 0;
+          mouseDev.has_x = 0;
+          mouseDev.has_y = 0;
+          break;
+        }
 	default: break;
       }
     } else if (evt->value == 0) { /* button up */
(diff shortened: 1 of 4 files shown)


krono:
Oh thats neat!

eliotmiranda:
Hi Tony,

   well that’s great because there *is* a better way available to test. The threaded FFI VM (see the various [gcc3x-]cointerpmt.c files) is currently being used in anger at LabWare. There is a functional threaded ODBC connect and an in progress socket library. I believe both will be released open source.

So the possibility exists to interface directly to the event system using the FFI and callbacks since one can now make a blocking call for the next event in one thread while the VM continues to execute Smalltalk in other threads.

Currently build setups exist only for macOS. But it’s simple. One uses [gcc3x-]cointerpmt.c in place of [gcc3x-]cointerp.c and defines COGMT=1 on the c compiler command line. One would add buildmt, buildastmt & builddbgmt directories alongside build et al.

The basic scheme is that FFI calls block the entire vm until the next heartbeat (currently 500hz, but could be higher freq) at which time the vm spawns another thread to run Smalltalk if blocked in an FFI call. The FFI thread on its return path and any new thread race to control the vm. Above that one can bind Smalltalk processes to threads using the threadAffinity inst var. The VM’s thread scheduler will switch threads if necessary in process switch.  There are a few more details.

So the threaded vm provides the necessary infrastructure to connect Smalltalk directly to the GUI subsystem.

platforms/unix/vm-display-fbdev/sqUnixEvdevKeyMouse.c
```diff
@@ -106,13 +110,39 @@ static void setSqueakModifierState();
...
+  int abs_type;         /* 0 => simple (singletouch), 1 => type A, 2 => type B */
+  int slots[NUM_SUPPORTED_SLOTS];
+  int primary_tracking_id; /* -1 for none */
+  int current_slot;     /* -1 for none */
+
+  double tap_release_deadline;
+  int tap_bit;
 };
 
+#define TAP_START_TIMEOUT 0.18
```
LinqLover, line 143: Would it make sense to deliver the raw touch events from the plugin to the image and interpret them there as mouse events? That way the programmer in the image would have maximum control and you would not need to hardcode constants in the VM. Sorry if my question does not make sense, I have only briefly skimmed the diff.

tonyg: Absolutely it would, but that would be a much more invasive change to the VM infrastructure than I have time for right now. On the other hand, I have already had good results from doing all this directly within the image: see package LinuxIO at https://squeaksource.com/SqueakPhone/, which does finger tracking etc by reading `/dev/input/event*` devices from within the image.

tonyg: The hardcoding of constants is something to address though. Suggestions welcome (yet more command-line arguments??).

tonyg:
> So the threaded vm provides the necessary infrastructure to connect Smalltalk directly to the GUI subsystem.

That is wonderful news! I'll look forward to it. However, the VM *today* is in good enough shape to allow the image to reach out and use `/dev/input/event*` on its own - see package LinuxIO at https://squeaksource.com/SqueakPhone/, which does just this.

This PR (still, I suppose, WIP) is instead intended to allow current Trunk images to work on framebuffer systems with a trackpad. For that, alterations to the C are both necessary and evil...

platforms/unix/vm-display-fbdev/sqUnixEvdevKeyMouse.c
eliotmiranda: To Tony & Christoph on the issue of hardcoded constants and no time to get the changes into the vm right now,  agree with both of you. Let’s approve the pull request, add an issue to remind us to provide primitive access to the tap time constant, and address it when time allows.

PR #744, opened 2025-10-26
arm64 OpenBSD

KenDickey:
Tested on RasPi4 + OpenBSD 7.8


KenDickey:
Updated x86 mvm's.
Tested good on x86 both with OpenBSD and Ubuntu Linux.

PR #746, opened 2025-11-18
[ci] add extra monthly builds for older linux/glibc versions

### LINKED ISSUE #742: Restore support for older glibc versions in CI builds
For example, recent builds no longer work on Ubuntu 20 due to missing glibc version, which is a bit inconvenient. Ubuntu 20 Pro is still supported until 2030, so that's a valid platform IMHO.

LinqLover:
Closes #742

Also includes minor revisions for linux prepare scripts:

- extract `sudo` from prepare scripts because docker does not need it (and specifying env vars like `DEBIAN_FRONTEND` from the outside is not possible if the script uses `sudo` without `-E`)
- add missing dev package `libltdl-dev` (seems to be preinstalled in GitHub runner but not a stock build-essentials Ubuntu image)
- ensure `add-apt-repository` is installed before using it for linux32x68 builds

**When merging, please consider doing a squash merge to keep the ancestry clean.**

### DIFF
--- a/.github/workflows/extra-linux-classic.yml
+++ b/.github/workflows/extra-linux-classic.yml
@@ -1,0 +1,111 @@
+name: Extra Monthly Builds (Linux Classic)
+# For older glibc version
+# See: https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/742
+
+on:
+  schedule:
+    - cron: '0 0 1 * *' # Build on the 1st of every month at midnight
+  workflow_dispatch:
+    inputs:
+      should-deploy:
+        description: 'Deploy on success'
+        required: true
+        default: "true" # No boolean support at the moment
+
+# !! NOTICE !!
+# This workflow spec is basically a copy of linux.yml. It includes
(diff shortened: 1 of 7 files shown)


LinqLover:
@marceltaeumel Does this look okay, can we merge it? :)

@dtlewis290 If you need it, you can already download a working VM for GLIBC 2.31 from [here](https://github.com/LinqLover/opensmalltalk-vm/actions/workflows/extra-linux-classic.yml) :-)

dtlewis290:
Excellent. I can confirm that your build artifact runs successfully on the older Linux image that we are using on squeaksource.com. This means that it resolves the glibc version issue on that Linux version 5.10.0-14-amd64 platform. This is a an older but still supported Linux with about 12 months remaining on its LTS lifespan, so it is likely typical of Linux versions that many Squeak end users will be using. I have so far tested only "./squeak -version" to confirm that the VM runs without link errors, and I will plan to do a test for real on the squeaksource.com server later on (maybe this weekend) when I can watch it and roll bock the change if any issues are seen.

PR #748, opened 2025-12-16
[ci] fix maunal workflow trigger for extra-linux-classic

LinqLover:


dtlewis290 (approved):
Presumed good update, let's give it a try

LinqLover:
@dtlewis290 Thank you for testing! For a more insightful check of this PR, you could also trigger another manual workflow run from here: https://github.com/OpenSmalltalk/opensmalltalk-vm/actions/workflows/extra-linux-classic.yml

I think Marcel wanted at some point that he wanted to build a third VM release candidate, but we better ask himself :D

PR #749, opened 2025-12-21
Add PseudoTTYPlugin to Linux and Solaris builds. Tested on Linux 64x6…

dtlewis290:
Tested on Linux 64x64 so far but this plugin can reasonably be expected to work on any Unix platform. To test, load Telnet301 from www.squeaksource.com/IanPiumartaGoodies repository, then evaluate "TeletypeWindow open" to start a terminal emulator and use its window menu to select "new shell session ..." to start a pty connected terminal session. Try running the Linux "top" command for example.

### DIFF
--- a/building/linux32/squeak.stack.spur/plugins.ext
+++ b/building/linux32/squeak.stack.spur/plugins.ext
@@ -16,4 +16,6 @@ XDisplayControlPlugin \
 DESPlugin \
 MD5Plugin \
 SHA2Plugin \
-VectorEnginePlugin 
+VectorEnginePlugin \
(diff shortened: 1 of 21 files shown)


dtlewis290:
There is one build gate failure, but it  appears to be unrelated to this change. The error happened when trying to fetch http://archive.raspbian.org/raspbian/pool/main/g/glib2.0/libglib2.0-dev_2.66.8-1%2bdeb11u7_armhf.deb which is blocked when I try to fetch from a browser but is accessible with https, so it might be just a matter of changing from http to https in a build script.

PR #753, opened 2026-01-09
Support for OpenBSD

KenDickey:
OpenBSD 7.8

Built & tested squeak.cog.spur/build/mvm on both x86_64 and Aarch64 on OpenBSD.
Also built & tested Aarch64 on Linux [Raspian,RasPi5].


dtlewis290 (approved):
These changes are specific to OpenBSD and appear to present no risk to other platforms. If Ken says it works on OpenBSD and we can see that it does not impact other platforms, then I recommend merging.

PR #764, opened 2026-02-25
Adaptive heartbeat + delay-aware idle sleep for macOS and Linux

johnmci:
The Squeak/Cuis VM idles at ~8% CPU even when no Smalltalk code is running. Two causes:

1. **The image requests 1ms sleep** via `relinquishProcessorForMicroseconds:` — the VM honors this literally, causing ~850 wakeups/sec
2. **The heartbeat thread runs at full speed** (2ms interval) even when the main thread is blocked in `select()`/sleep, delivering unnecessary timer ticks

### Phase 4: Adaptive heartbeat (macOS)

- Add `mainThreadIsIdle` flag, set by main thread before sleeping, cleared on wake
- When idle, heartbeat backs off to 50ms (`IDLE_BEAT_MS`) — still accurate enough for Delay resolution
- Replace `select()` with **kqueue** on macOS (`USE_KQUEUE`/`__APPLE__` gated) — consolidates heartbeat timer, AIO file descriptors, and display pipe into a single `kevent()` call
- When idle, fold the heartbeat timer into the kqueue wait — one syscall instead of two threads polling

### Phase 5: Delay-aware idle sleep (Linux)

- Port the `getNextWakeupUsecs()` logic to Linux X11 and null display drivers
- Ignore the image's requested sleep time; instead sleep for `min(timeUntilNextDelay, 500ms)`
- If a Delay is already overdue, return immediately (no sleep)
- Set `mainThreadIsIdle` with memory barriers (`__sync_synchronize`) for heartbeat throttling

## Files changed

- `platforms/unix/vm-display-X11/sqUnixX11.c` — delay-aware idle for X11
- `platforms/unix/vm-display-null/sqUnixDisplayNull.c` — delay-aware idle for headless
- `platforms/unix/vm/sqUnixHeartbeat.c` — `mainThreadIsIdle` made non-static; adaptive heartbeat
- `platforms/iOS/vm/Common/sqMacV2Time.c` — kqueue + adaptive heartbeat (macOS)

## Results

| Platform | Before | After | Reduction |
|----------|--------|-------|-----------|
| macOS (kqueue) | ~8-9% | ~0.8% | 91% |
| Linux x86_64 | 7.7% | 1.1% | 86% |

Tested with Squeak 6.0 and Cuis on both platforms. No image-side changes required — all improvements are VM-level.

## Design notes

- All macOS/kqueue changes gated on `USE_KQUEUE`/`__APPLE__` — Linux epoll path untouched
- `MAX_IDLE_USECS` set to 500ms after benchmarking 50ms/500ms/1000ms — best balance of CPU savings vs Delay accuracy
- Heartbeat thread is still needed for preemption of active Smalltalk code — only throttled when idle
- timerfd/epoll consolidation on Linux deferred to future work

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -4858,10 +4858,38 @@ display_ioBeep(void)
 }
 
 
+#define MAX_IDLE_USECS 500000  /* 500ms cap when no Delays pending */
+
 static sqInt
 display_ioRelinquishProcessorForMicroseconds(sqInt microSeconds)
 {
-  aioSleepForUsecs(handleEvents() ? 0 : microSeconds);
+  extern usqLong getNextWakeupUsecs(void);
+  extern volatile int mainThreadIsIdle;
+
+  if (handleEvents())
+    return 0;
+
+  usqLong nextWakeupUsecs = getNextWakeupUsecs();
+  usqLong utcNow = ioUTCMicroseconds();
+  long realTimeToWait;
+
+  if (nextWakeupUsecs != 0 && nextWakeupUsecs <= utcNow)
+    return 0;  /* Delay already overdue, don't sleep */
+
+  if (nextWakeupUsecs != 0) {
+    realTimeToWait = nextWakeupUsecs - utcNow;
+    if (realTimeToWait > MAX_IDLE_USECS)
+      realTimeToWait = MAX_IDLE_USECS;
+  } else {
+    realTimeToWait = MAX_IDLE_USECS;
+  }
+
+  mainThreadIsIdle = 1;
+  __sync_synchronize();
+  aioSleepForUsecs(realTimeToWait);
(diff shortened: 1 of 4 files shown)


LinqLover:
That sounds very interesting. Only recently I noticed that the energy consumption of Squeak on my Linux laptop is comparably very high due to many wake-ups (2-3 W per image according to powertop). It would be interesting to learn how much this change can reduce energy consumption.

> heartbeat backs off to 50ms (`IDLE_BEAT_MS`) — still accurate enough for Delay resolution

Is it? In Squeak we documented the expectation of ~1 ms precision in `Duration>>wait`. With 50 times as much, I wonder whether any code such as concurrent animations or other high-frequency loops (not necessarily in the standard images) might stop working as expected. Any packages that uses `1 milliSecond wait` 1000 times within a longer operation would suddenly display degraded performance.

Should this constant perhaps be customizable through a VM parameter from the image side? Just brainstorming. This looks quite interesting.

PR #768, opened 2026-04-26
SqueakSSL: Drain pending handshake data after connection success (required for TLS 1.3)

LinqLover:
The LibreSSL implementation already does this. Windows and macOS versions of the plugin are currently pinned to TLS 1.2 but can benefit from this patch in the future, too.

On the image side, this change means that code must no longer rely on primitiveConnect returning 0 for testing whether the connection has established but test SQSSL_PROP_SSLSTATE analogously to for primitiveAccept. While this is a breaking change, it is one that was already dictated by TLS 1.3 in the past few years and has caused a red bar on Linux. Platforms that do not use TLS 1.3 (Windows, macOS) are not affected by this change.

Will be complemented by [SqueakSSL-Core-ct.37](https://source.squeak.org/inbox/SqueakSSL-Core-ct.37.diff)/[SqueakSSL-Tests-ct.29](https://source.squeak.org/inbox/SqueakSSL-Tests-ct.29.diff) (Squeak inbox).

### DIFF
--- a/platforms/iOS/plugins/SqueakSSL/sqMacSSL.c
+++ b/platforms/iOS/plugins/SqueakSSL/sqMacSSL.c
@@ -550,7 +550,7 @@ sqInt sqConnectSSL(sqInt handle, char* srcBuf, sqInt srcLen, char* dstBuf,
     if (status == noErr) {
         sqExtractPeerName(ssl);
     }
-    return SQSSL_OK;
+    return ssl->outLen;
 }
 
 /* sqAcceptSSL: Start/continue an SSL server handshake.
@@ -613,7 +613,7 @@ sqInt sqAcceptSSL(sqInt handle, char* srcBuf, sqInt srcLen, char* dstBuf,
     }
     /* We are connected. Verify the cert. */
     ssl->state = SQSSL_CONNECTED;
-        return SQSSL_OK;
+    return ssl->outLen;
 }
 
 /* sqEncryptSSL: Encrypt data for SSL transmission.
--- a/platforms/unix/plugins/SqueakSSL/sqOBSDOpenSSL.inc
+++ b/platforms/unix/plugins/SqueakSSL/sqOBSDOpenSSL.inc
@@ -498,7 +498,7 @@ sqInt sqConnectSSL(sqInt handle, char* srcBuf, sqInt srcLen, char *dstBuf, sqInt
 	} else {
 		ssl->certFlags = SQSSL_NO_CERTIFICATE;
 	}
-	return 0;
+	return sqCopyBioSSL(ssl, ssl->bioWrite, dstBuf, dstLen);
 }
 
 /* sqAcceptSSL: Start/continue an SSL server handshake.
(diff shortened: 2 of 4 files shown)


LinqLover:
:warning: Before merging this change, it should be tested on Windows/macOS/OBSD, too!

krono:
This definitely would need an increment in the Version number.

LinqLover:
Why? This does not introduce any backward compatibility per se. For older TLS versions, primitiveConnect keeps returning 0. If at all, we should have introduced a new version number when adding support for TLS 1.3 to the Unix VM.

krono:
It changes the contract. If the Image-side needs changes due to chages in the plugin, the version must be updated.

krono:
Where does copilot come from?

LinqLover:
> It changes the contract. If the Image-side needs changes due to chages in the plugin, the version must be updated.

Okay, I see your point. The contract was effectively already changed by 2018 when OpenSSL supported OpenSSL 1.1.1 and our unix plugin started automatically using it. Since then, the change in the contract `primitiveConnect` has been that `primitiveConnect` would not drain all data from the handshake, leaving callers that control the stream themselves with an incomplete handshake. The change in this PR restores the previous expectation of draining the full handshake by doing an alternative contract change and returning something else than just `0` from `primitiveConnect`. So, yes, if you want to, we can increase the version (I don't know how). But I don't think that the image code needs to check this version for a fallback.

LinqLover:
> Where does copilot come from?

From my GitHub account, I'm still testing this (a number of times it already provided me C noob with helpful comments, though not too often). It's not a setting of this repository.

krono:
I have not consented my code to be subjected to copilot.

johnmci:
As far as I remember code contributions to squeak or the open source vm require the author to agree to https://github.com/OpenSmalltalk/opensmalltalk-vm/blob/Cog/LICENSE

Which doesn't allow the author to restrict how or who or what examines his code

krono:
I would have expected this to be discussed (eg mainling list) instead of being silently applied.

LinqLover:
@krono @johnmci I have started a new issue https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/769 for this where we can discuss the use of GitHub Copilot Reviews without further polluting the comments of this PR. :-)

krono:
ok

LinqLover:
> > It changes the contract. If the Image-side needs changes due to chages in the plugin, the version must be updated.
> 
> Okay, I see your point. The contract was effectively already changed by 2018 when OpenSSL supported OpenSSL 1.1.1 and our unix plugin started automatically using it. Since then, the change in the contract `primitiveConnect` has been that `primitiveConnect` would not drain all data from the handshake, leaving callers that control the stream themselves with an incomplete handshake. The change in this PR restores the previous expectation of draining the full handshake by doing an alternative contract change and returning something else than just `0` from `primitiveConnect`. So, yes, if you want to, we can increase the version (I don't know how). But I don't think that the image code needs to check this version for a fallback.

Please ignore the above. I double-checked. With this patch, a slightly older trunk image cannot establish a SSL connection because the previous termination condition `result = 0` no longer works. Is this the kind of backward compatibility that we are interested in or do we only care about the other direction (older VM, newer image)? I'm always confused. If compatibility matters here, what would be the best way to preserve that?

1. Make the change backward-compatible by making `sqConnectSSL` return `0` if already connected (not necessarily good style)?
2. Only activate the new behavior if the image version is newer than X or a VM flag is set?
3. Add a second version of the primitive with the new behavior and keep the old one for compatibility?
4. others?

No idea how this kind of thing is typically done in the VM. Advice welcome!

PR #775, opened 2026-06-27
Fix Linux clipboard when running the VM in headless mode

LinqLover:
Previously, `squeak --headless squeak.image --evaluate "Clipboard clipboardText"` led to a segfault. Without a connection to the X server, the clipboard is not available, so do not attempt accessing it.

### DIFF
--- a/platforms/unix/vm-display-X11/sqUnixX11.c
+++ b/platforms/unix/vm-display-X11/sqUnixX11.c
@@ -1231,6 +1231,9 @@ static Atom stringToAtom(char *target, size_t size)
 static void
 display_clipboardWriteWithType(char *data, size_t ndata, char *typeName, size_t nTypeName, int isDnd, int isClaiming)
 {
+  if (!isConnectedToXServer)
+    return;
+
   if (allocateSelectionBuffer(ndata))
     {
       Atom type= stringToAtom(typeName, nTypeName);
@@ -1246,6 +1249,9 @@ display_clipboardWriteWithType(char *data, size_t ndata, char *typeName, size_t 
 static sqInt
 display_clipboardSize(void)
 {
+  if (!isConnectedToXServer)
+    return 0;
+
   if (stOwnsClipboard) return 0;
   getSelection();
   return stPrimarySelectionSize;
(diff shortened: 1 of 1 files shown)


marceltaeumel:
I wonder if your assumption is correct here.

```C
if (!isConnectedToXServer)
    return 0;
```

Is this the typical pattern in `sqUnixX11.c`? Why would you go through that module if you do not have an X server? There was some other way to avoid graphical output and be "headless" ... Hmm...

LinqLover:
> I wonder if your assumption is correct here.
> 
> ```c
> if (!isConnectedToXServer)
>     return 0;
> ```
> 
> Is this the typical pattern in `sqUnixX11.c`? Why would you go through that module if you do not have an X server? There was some other way to avoid graphical output and be "headless" ... Hmm...

We find the same pattern in `platforms/unix/vm-display-null/sqUnixDisplayNull.c`, so the contract is established in the VM.

The alternative to `--headless` (`sqUnixX11.c`) on Linux is `-vm-display-null`/`-nodisplay` (`sqUnixDisplayNull.c`). As I understand it, both behave pretty much equivalently. Using `xqUnixX11`, headless mode can also enabled dynamically via `primitiveDisconnectDisplay`, which is used by OSProcess to fork a headless child.

But the general pattern of defining a fallback for headless mode/missing X server is not new to `sqUnixX11.c`. For instance, getting/setting the scale factor, screen size, or cursor does return similar defaults...

eliotmiranda:
Is it possible that --headless would select -vm-display-null at startup Then these checks wouldn't be necessary.

dtlewis290:
For background: The original motivation for isConnectedToXServer was to support XDisplayControlPlugin, which in turn is used to support the various forkSqueak methods in OSProcess. It is implemented in the X11 display module. Conceptually,  X11 is independent of unix, and unix can have other kinds of displays, so doing X11 related things in the X11 display module is a good thing to do.

The unix clipboard plugin already has some unfortunate direct dependencies on X11, but that issue is beyond the scope of this pull request.

My opinion: This pull request is simple and harmless. It is putting the new logic into the X11 module where it belongs, so it looks good to me.

LinqLover:
@eliotmiranda I'd support Dave's perspective. `-vm-display-null` would fix the issue for the static case, but dynamic changes to the X connection via primitiveDisconnectDisplay would still fail.

dtlewis290:
Are there any objections to merging this pull request? If not I will merge it in a couple of days or so. Thanks.

dtlewis290:
Having heard no objections it is time to merge. Original branch is stale and the changes are small so I committed them directly with credit to Christoph in commit https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/f2829eaadde0bb5bb45d339d6233e4ae993408dd

PR #778, opened 2026-07-29
Add SqueakLiveTyping VM build process (ARM + x64 on Win/Linux/Mac)

hernanwilkinson:
## What

Adds a first-class, CI-driven build for the **LiveTyping** variant of the Spur 64-bit Cog VM, branded **SqueakLiveTyping**, coexisting with the stock Squeak VM. Produced for the six modern 64-bit targets: `macos64x64`, `macos64ARMv8`, `win64x64`, `win64ARMv8`, `linux64x64`, `linux64ARMv8`.

## Why

LiveTyping is a runtime type-collection feature baked into a regenerated Spur64 Cog source tree (`LiveTypingCoInterpreter` / `LiveTypingCogit`, VMMaker.oscog-eem.3776). Until now it had no dedicated artifact or CI. This PR turns it into a proper, reproducible build flavor.

## How

- **`src/spur64.cog.livetyping/`** — dedicated generated source tree, split out from the stock `src/spur64.cog` so both VMs can be built side by side.
- **`building/<arch>/squeak.cog.spur.livetyping/`** — a new flavor dir per target, mirroring each platform's `squeak.cog.spur` but pointing `VMSRCDIR` / `--with-src` at the livetyping tree and branding the artifact **SqueakLiveTyping**:
  - macOS: new `common/Makefile.app.squeaklivetyping` sets `APPNAME:=SqueakLiveTyping` -> `SqueakLiveTyping.app`
  - Windows: `VM:=SqueakLiveTyping` + `SqueakLiveTyping.def.in` / `.rc` / `.exe.manifest`
  - Linux: `build*/mvm` use `--with-src=src/spur64.cog.livetyping`, `INSTALLDIR=sqcogspurlt64...`
- **CI** — `squeak.cog.spur.livetyping` added to `matrix.flavor` of the six platform workflows, with `exclude` blocks so it builds only the supported 64-bit (and threaded-heartbeat on Linux) legs. It rides the existing `actions_build.sh` -> `pack-vm.sh` -> release pipeline, producing `squeak.cog.spur.livetyping_<arch>` artifacts.

## Verification

Local `macos64ARMv8` fast build produces `SqueakLiveTyping.app` whose binary reports `LiveTypingCoInterpreter` / `LiveTypingCogit`. The stock `squeak.cog.spur` build is unaffected (it still compiles the restored stock `src/spur64.cog`).

## Note

The regenerated LiveTyping `interp.h` does not `#define IMMUTABILITY 1` (immutability OFF), as generated by VMMaker; no build config restores it. Flagging in case that should change.


eliotmiranda (changes requested):
Hi Hernan,

    the name scheme for the build directories should be
        squeak.livetyping.spur
This matches e.g. squeak.sista.spur, since the live typing machinery implies cog. Sorry to ask you to do this but please resubmit the pull request with all squeak.cog.spur.livetyping directories renamed to squeak.livetyping.spur

hernanwilkinson:
@eliotmiranda I renamed the directory as you suggested, and changed the related files.
Let me know if it is ok or if I should change something else.

platforms/win32/plugins/SqueakFFIPrims/winARM64Thunk.c
marceltaeumel, line 1: Have you tried the regular OSVM builds? Callbacks (i.e. thunks) work fine on Windows ARM64. This change seems unnecessary unless your build works differents compared to win64ARMv8/squeak.cog.spur.

https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/download/latest-build/squeak.cog.spur_win64ARMv8.zip

marceltaeumel: If you are referring to a MinGW-w64 toolchain, please expand your commentary. I am referring to WinSDK builds, which we do in CI.

platforms/win32/vm/sqWin32Backtrace.c
marceltaeumel, line 1: This is a nice contribution for an extra pull request. It is rather surprising in this one. Please add also some more information about the change.

platforms/win32/vm/sqWin32Heartbeat.c
marceltaeumel, line 1: This is a nice contribution for an extra pull request. It is rather surprising in this one. Please add also some more information about the change.

marceltaeumel:
I object to the changes in the following files, which should rather be proposed as an extra pull request.

```
platforms/win32/plugins/SqueakFFIPrims/Makefile.plugin
platforms/win32/plugins/SqueakFFIPrims/winARM64Thunk.c
platforms/win32/vm/sqWin32Backtrace.c
platforms/win32/vm/sqWin32Heartbeat.c
```

As I commented in `winARM64Thunk.c`, I am unsure whether that change is necessary at all as it works already without in our regular builds.

